PR #12, opened 2015-10-02
Report physical blocks used

andrewgaul:
Tools like `du` use this to calculate the actual space used, accounting
for sparse files.

### DIFF
--- a/samples/statfs/statfs_test.go
+++ b/samples/statfs/statfs_test.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
+	"syscall"
 	"testing"
 
 	"github.com/jacobsa/fuse/fuseops"
@@ -203,4 +204,25 @@ func (t *StatFSTest) WriteSize() {
 	default:
 		AddFailure("Unhandled OS: %s", runtime.GOOS)
 	}
+}
+
+func (t *StatFSTest) StatBlocks() {
+	var err error
+	var stat syscall.Stat_t
+	const fileName = "foo"
+	const size = 1 << 22
+
+	err = ioutil.WriteFile(
+		path.Join(t.Dir, fileName),
+		bytes.Repeat([]byte{'x'}, size),
+		0400)
+	AssertEq(nil, err)
+
+	t.fs.SetStatResponse(fuseops.InodeAttributes{
+		Size: size,
+	})
+
+	err = syscall.Stat(path.Join(t.Dir, fileName), &stat)
+	AssertEq(nil, err)
+	ExpectEq(size/512, stat.Blocks)
 }
--- a/conversions.go
+++ b/conversions.go
@@ -633,6 +633,8 @@ func convertAttributes(
 	out.Nlink = in.Nlink
 	out.Uid = in.Uid
 	out.Gid = in.Gid
+	// round up to the nearest 512 boundary
+	out.Blocks = (in.Size + 512 - 1) / 512
 
 	// Set the mode.
 	out.Mode = uint32(in.Mode) & 0777
--- a/samples/statfs/statfs.go
+++ b/samples/statfs/statfs.go
@@ -40,6 +40,9 @@ type FS interface {
 	// Set the canned response to be used for future statfs ops.
 	SetStatFSResponse(r fuseops.StatFSOp)
 
+	// Set the canned response to be used for future stat ops.
+	SetStatResponse(r fuseops.InodeAttributes)
+
 	// Return the size of the most recent write delivered by the kernel, or -1 if
 	// none.
 	MostRecentWriteSize() int
@@ -47,6 +50,9 @@ type FS interface {
 
 func New() (fs FS) {
 	fs = &statFS{
+		cannedStatResponse: fuseops.InodeAttributes{
+			Mode: 0666,
+		},
 		mostRecentWriteSize: -1,
 	}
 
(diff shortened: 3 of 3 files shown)


jacobsa:
Very sorry for the slow response; I've been travelling. Could you please say a bit about why you need this, and add integration tests for Linux and OS X in [this](https://github.com/jacobsa/fuse/tree/master/samples/statfs) package? The cross-platform behavior of `statfs(2)` is quite weird, and I figure this might be similar.

andrewgaul:
I amended the commit with a more complete description and a unit test.

jacobsa:
I'm still interested to hear _why_ you need this. Are you writing a fuse file system where physical blocks are relevant?

samples/memfs/inode.go
```diff
@@ -331,6 +332,8 @@ func (in *inode) WriteAt(p []byte, off int64) (n int, err error) {
 		padding := make([]byte, newLen-len(in.contents))
 		in.contents = append(in.contents, padding...)
 		in.attrs.Size = uint64(newLen)
+		var rounded = (uint64(newLen) + uint64(in.attrs.BlockSize) - 1) & ^uint64(in.attrs.BlockSize-1)
```
jacobsa: Please revert the changes to memfs, and put the tests in [statfs](https://github.com/jacobsa/fuse/blob/master/samples/statfs/statfs.go), where the existing tests for related functionality of `statfs(2)` live. You'll want to add a method called `SetBlockInfo` or similar that affects the output of `GetInodeAttributes` for `childInodeID`.

Don't forget to add tests for what happens when each value is left at its default.

fuseops/simple_types.go
```diff
@@ -95,6 +95,12 @@ type InodeAttributes struct {
 	// Ownership information
 	Uid uint32
 	Gid uint32
+
+	// The number of blocks allocated to the file, 512-byte units. (This may be
+	// smaller than Size/512 when the file has holes.)
+	Blocks uint64
+	// The "preferred" blocksize for efficient filesystem I/O.
```
jacobsa: Blank line before this.

jacobsa: "file system"

fuseops/simple_types.go
```diff
@@ -95,6 +95,12 @@ type InodeAttributes struct {
...
+	BlockSize uint32
```
jacobsa: I know that you're following the convention used by `stat(2)`, but it's quite confusing for this to be called "block size" when the unit of `Blocks` is already defined to be 512 bytes. Please name this `IoSize` for symmetry with [`StatFSOp`](https://godoc.org/github.com/jacobsa/fuse/fuseops#StatFSOp).

jacobsa:
Ping

andrewgaul:
> I'm still interested to hear why you need this. Are you writing a fuse file system where physical blocks are relevant?

Several commands like `du` and the first line from `ls -l` report the physical size used, not the logical size.  The file system I use, https://github.com/kahing/goofys, does not support sparse files and will set the two sizes to the same value.

fuseops/simple_types.go
andrewgaul: Done.

fuseops/simple_types.go
andrewgaul: Done.

samples/memfs/inode.go
andrewgaul: Done.

andrewgaul:
Addressed all review comments.

jacobsa:
Thanks for your patience. I have a few more questions:

1.  Your test doesn't pass on OS X, where there seem to be two differences from Linux:

    * osxfuse uses an older protocol where `fuse_attr::blksize` is not communicated to the kernel. `stat(2)` seems to always return 1 MiB for `stat::st_blksize`.
    * OS X appears to ignore the value of `fuse_attr::blocks` and always return a value of `stat::st_blocks` that is slightly greater than `stat::st_size` divided by 512.
    
    I'm happy to fix the tests, but is this behavioral difference a problem for you?

2.  I now understand that the problem is that `du mount_point/foo` shows zero on Linux, no matter the reported size of `foo`. (This is not true on OS X, because of the above.) Would it be good enough for your purposes if this package automatically set `fuse_attr::blocks` to `ceil(size / 512)`? That would give nearly-consistent behavior across operating systems, without exposing a knob that is confusing because it doesn't work on OS X.

3.  I don't see how setting the preferred I/O size (`stat::st_blksize`) is motivated by the reason you gave [above](https://github.com/jacobsa/fuse/pull/12#issuecomment-149378749)? Forgive my curiosity; I'm keen to understand how this comes up in practice.

andrewgaul:
Interesting about the differences between Mac OS X and Linux FUSE.  I will simplify this with 512-byte blocks.

samples/statfs/statfs_test.go
```diff
@@ -204,3 +205,24 @@ func (t *StatFSTest) WriteSize() {
 		AddFailure("Unhandled OS: %s", runtime.GOOS)
 	}
 }
+
+func (t *StatFSTest) BlocksAndBlockSize() {
```
jacobsa: Should be renamed now; probably `StatBlocks`.

jacobsa:
Thanks! Do you mind fixing the commit message before I pull?

samples/statfs/statfs_test.go
andrewgaul: Done.

andrewgaul:
Updated commit message.

PR #18, opened 2016-09-25
Support OSXFUSE 3.x, fall back to 2.x if needed

zbiljic:
Also supports custom bundled OSXFUSE installs.

The homebrew cask osxfuse version is now 3.5.0: https://github.com/caskroom/homebrew-cask/blob/master/Casks/osxfuse.rb#L2.

This code is basically transferred from bazil.org/fuse.

### DIFF
--- a/mount_config.go
+++ b/mount_config.go
@@ -135,6 +135,14 @@ type MountConfig struct {
 	// default name involving the string 'osxfuse' is used.
 	VolumeName string
 
+	// OS X only.
+	//
+	// OSXFUSELocations sets where to look for OSXFUSE files. The arguments are
+	// all the possible locations. The previous locations are replaced.
+	//
+	// Without this option, OSXFUSELocationV3 and OSXFUSELocationV2 are used.
+	OSXFUSELocations []OSXFUSEPaths
+
 	// Additional key=value options to pass unadulterated to the underlying mount
 	// command. See `man 8 mount`, the fuse documentation, etc. for
 	// system-specific information.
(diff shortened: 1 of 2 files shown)


zbiljic:
First, I would like to apologize for delayed response.
Second, you are completly right in wanting to have references for these changes; stellar documentation is why I chose this library in the first place. I must have spend two days going through the commits just to verify everything that you are asking.

As for the answers:

1. All the variables are indeed needed.
2. Yes. They have changed the osxfuse file system library paths and options, even during development of osxfuse 3.x.
3. At first `OSXFUSEPaths` were internal, but someone might (at some point) have the need for custom osxfuse build, and this way it is easy to do so.

I will attempt to provide updated pull request as soon as I am able to.
Also, if you think some things should be changed, I won't mind to do so.

jacobsa:
Thanks! Sorry for _my_ slow response; I've been away on holiday. As for what needs to be done here:

1. Could you please make all of the new symbols unexported again? We can export them later if a need truly arises. I would rather have the library just do the right thing.
2. It sounds like you did a bunch of research on this. Could you please link (in the code comments) to your sources?

It would be best if you could do these as new commits in the pull request rather than overwriting the existing one. I may wind up pulling this as-is if osxfuse 3 support becomes urgent, so would like to be able to cleanly make these fixes afterward.

zbiljic:
Just as I started to commit changes, you already finished.
Sorry again, this was really the only free time I've had in previous weeks.

jacobsa:
No worries, thanks for the fix. I'm still very interested in getting references from you if you have them.

PR #23, opened 2017-03-09
Fix build on ARM, and revert old workaround (not needed anymore)

elProxy:


internal/buffer/out_message.go
```diff
@@ -55,22 +55,8 @@ func init() {
 // Reset resets m so that it's ready to be used again. Afterward, the contents
 // are solely a zeroed fusekernel.OutHeader struct.
 func (m *OutMessage) Reset() {
-	// Ideally we'd like to write:
-	//
-	//     m.payloadOffset = 0
-	//     m.header = fusekernel.OutHeader{}
-	//
-	// But Go 1.8 beta 2 generates bad code for this
```
jacobsa: Have you verified that Go 1.8 generates decent code here after reverting this? Please link to a gist with before and after disassembly in your commit description.

Hitechcomputergeek:
Please merge the second patch. Build still fails on ARM, and and applying the second patch seems to fix it.

PR #25, opened 2017-05-04
Add direct-io option to OpenFileOp

bsidhom:
The current API does not support opening files for direct-io.

When enabled, this ensures that all user space calls to read and write actually make it to the fuse layer.

Having this option is useful for filesystems whose file sizes are dynamic.

### DIFF
--- a/fuseops/ops.go
+++ b/fuseops/ops.go
@@ -601,8 +601,7 @@ type ReadFileOp struct {
 	// where EOF is by checking the inode size (http://goo.gl/0BkqKD), returned
 	// by a previous call to LookUpInode, GetInodeAttributes, etc.
 	//
-	// If direct io is enabled, semantics should match those of read (man 2
-	// read).
+	// If direct IO is enabled, semantics should match those of read(2).
 	BytesRead int
 }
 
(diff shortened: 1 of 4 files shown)


fuseops/ops.go
```diff
@@ -591,6 +600,9 @@ type ReadFileOp struct {
 	// (http://goo.gl/SGxnaN) to read a page at a time. It appears to understand
 	// where EOF is by checking the inode size (http://goo.gl/0BkqKD), returned
 	// by a previous call to LookUpInode, GetInodeAttributes, etc.
+	//
+	// If direct io is enabled, semantics should match those of read (man 2
```
jacobsa: "direct IO"

bsidhom:
I've added dyamic_fs to test this. I'm not sure that name is appropriate, so let me know if you have any suggestions.

jacobsa:
Thanks! Unfortunately the test fails on macOS Sierra. Travis isn't configured to use it, but the output from my machine is [here](https://gist.githubusercontent.com/jacobsa/c0ba9026b7e1cc1005a6d26f73d8aa88/raw/3990c45f7ad24308c2bf408c65c46d72b68d9514/gistfile1.txt). Could you please take a look?

bsidhom:
Looks like Mac invokes statfs on os.Lstat/os.Stat. Should work now.

jacobsa:
Rebased and merged in 77e8f7f7670aade0e3f76295ec3dd9a34ed01874. Thank you!

PR #27, opened 2017-05-18
Fix FTBFS on i386/armel/armhf platform

rogers0:
Original Debian Bug:
  https://bugs.debian.org/860703

### DIFF
--- a/fusetesting/stat_linux.go
+++ b/fusetesting/stat_linux.go
@@ -30,7 +30,7 @@ func extractBirthtime(sys interface{}) (birthtime time.Time, ok bool) {
 }
 
 func extractNlink(sys interface{}) (nlink uint64, ok bool) {
-	nlink = sys.(*syscall.Stat_t).Nlink
+	nlink = (uint64)(sys.(*syscall.Stat_t).Nlink)
 	ok = true
 	return
 }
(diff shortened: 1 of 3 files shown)


fusetesting/stat_linux.go
```diff
@@ -30,7 +30,7 @@ func extractBirthtime(sys interface{}) (birthtime time.Time, ok bool) {
 }
 
 func extractNlink(sys interface{}) (nlink uint64, ok bool) {
-	nlink = sys.(*syscall.Stat_t).Nlink
+	nlink = (uint64)(sys.(*syscall.Stat_t).Nlink)
```
jacobsa: Please remove parens: `uint64(foo)`

internal/buffer/runtime_go1.8.s
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build amd64 arm64 ppc64 ppc64le
+// +build 386 amd64 arm arm64 ppc64 ppc64le
```
jacobsa: Are you sure that the frame sizes listed below (`$0-16`, `$0-24`) are correct on these platforms? It seems like they're double what they should be...

internal/buffer/runtime_other.s
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build amd64 arm64 ppc64 ppc64le
+// +build 386 amd64 arm arm64 ppc64 ppc64le
```
jacobsa: Ditto here about frame sizes.

PR #30, opened 2017-10-06
handle forget inline instead of in goroutine

kahing:
when we are under memory pressure, or echo 3 > /proc/sys/vm/drop_caches,
kernel can send us many forget ops at the same time if we have lots of
inodes in memory. Running them all in goroutines can lead to even more
memory usage and eventually OOM.

### DIFF
--- a/fuseutil/file_system.go
+++ b/fuseutil/file_system.go
@@ -72,8 +72,10 @@ type FileSystem interface {
 // method.Respond with the resulting error. Unsupported ops are responded to
 // directly with ENOSYS.
 //
-// Each call to a FileSystem method is made on its own goroutine, and is free
-// to block.
+// Each call to a FileSystem method (except ForgetInode) is made on
+// its own goroutine, and is free to block. ForgetInode may be called
+// synchronously, and should not depend on calls to other methods
+// being received concurrently.
 //
 // (It is safe to naively process ops concurrently because the kernel
 // guarantees to serialize operations that the user expects to happen in order,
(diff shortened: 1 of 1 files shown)


jacobsa:
There are some potential performance and deadlock issues that I'm trying to
think through here:

* fileSystemServer.handleOp [calls into](https://github.com/kahing/fuse-1/blob/bf4c03d484a827d08b14edb1c56aa8ba59ab1e0b/fuseutil/file_system.go#L211) Connection.Reply, which [eventually](https://github.com/kahing/fuse-1/blob/bf4c03d484a827d08b14edb1c56aa8ba59ab1e0b/connection.go#L494) writes the response to the kernel  [https://github.com/kahing/fuse-1/blob/bf4c03d484a827d08b14edb1c56aa8ba59ab1e0b/connection.go#L490-L498](if kernelResponse returns false). Fortunately it [returns true](https://github.com/kahing/fuse-1/blob/bf4c03d484a827d08b14edb1c56aa8ba59ab1e0b/conversions.go#L551-L553) for ForgetInodeOp. So we don't have to worry about a deadlock where the kernel waits to consume the response until we've pulled more off the incoming queue, but we're blocking on writing the response. So I think this is okay.

* The only other substantive thing that `handleOp` does is  [call FileSystem.ForgetInode](https://github.com/kahing/fuse-1/blob/bf4c03d484a827d08b14edb1c56aa8ba59ab1e0b/fuseutil/file_system.go#L145). I worry that it may be an issue if this blocks on some other op coming in. But I can't think of any case where that would be reasonable, and indeed in the [gcsfuse implementation](https://github.com/GoogleCloudPlatform/gcsfuse/blob/7c92455a37c9901d64ffbb15caac1ba0de51fefe/internal/fs/fs.go#L1059-L1076) at least the function wouldn't behave any different with a synchronous call.

So I'm happy with this overall. Could you please make a few changes though?

1. Add comments in the code you've changed that say why you're calling synchronously in this case. Something like "Special case: call in this goroutine for forget inode ops, which may come in a flurry from the kernel and are generally cheap for the file system to handle."

2. Fix [the comment](https://github.com/kahing/fuse-1/blob/bf4c03d484a827d08b14edb1c56aa8ba59ab1e0b/fuseutil/file_system.go#L75-L76) on NewFileSystemServer that's no longer true. Document that ForgetInode may be called synchronously, and should not depend on calls to other methods being received concurrently.

kahing:
amended commit with comments

jacobsa:
Thanks!

PR #35, opened 2017-11-03
Add notification messages framework

inevity:
Now we  only impl the NotifyInvalInode,NotifyInvalEntry,
NotifyDelete.

Notify funcs allow the filesystem to invalidate VFS caches.

The statement that 'In a networked fuse file-system, the local
fuse instance may receive changes from a remote system that do not
pass through the local kernel vfs.
When those changes are made the local fuse instance must notify the
local kernel of the change so as to keep the local kernel's cache
in sync. This is accomplished via notify delete or notify
invalidate.
In other words, these notifications only apply if fuse file-system
makes changes to the file-system that are not initiated by the
local kernel: changes initiated by the local kernel must not
involve notify delete or notify invalidate (will deadlock), and
are not required.' is quoted from Jhon Muir.

About fuse  notify func history ,please ref link
http://lkml.iu.edu/hypermail/linux/kernel/0906.0/01120.html


connection.go
```diff
@@ -241,7 +241,7 @@ func (c *Connection) beginOp(
 	// should not record any state keyed on their ID.
 	//
 	// Cf. https://github.com/osxfuse/osxfuse/issues/208
-	if opCode != fusekernel.OpForget {
+	if opCode != fusekernel.OpForget && opCode < 100 {
```
jacobsa: Why this change?

connection.go
```diff
@@ -411,6 +411,36 @@ func (c *Connection) ReadOp() (ctx context.Context, op interface{}, err error) {
 	}
 }
 
+func (c *Connection) SetNotifyContext(op interface{}) (context.Context, error) {
+
```
jacobsa: Please remove the blank line here. (Please run gofmt if you haven't.)

connection.go
```diff
@@ -411,6 +411,36 @@ func (c *Connection) ReadOp() (ctx context.Context, op interface{}, err error) {
 	}
 }
 
+func (c *Connection) SetNotifyContext(op interface{}) (context.Context, error) {
```
jacobsa: Please add documentation for all new exported symbols.

fuseops/ops.go
```diff
@@ -845,3 +845,24 @@ type SetXattrOp struct {
 	// simply replace the value if the attribute exists.
 	Flags uint32
 }
+type NotifyPollOp struct {
```
jacobsa: It's especially important to have good documentation for the new Op structs. See the other ops in this package for an example of the kind of thing I'm looking for.

inevity:
Have added documentation.

connection.go
inevity: For notify ops ,it is issued by the userspace filesystem to
fuse kernel. The ops's kernel opCode is 0 in kernel' viewpoint or kernel not
care whether what the opcode is .But the ops have the notifycode  from 1
to 6 now. For adapt the notfiy ops to the BeginOP ,we fake an opcode as
notfiycode + 100. Now opcode range of ops from kernel including origin
 from userspace and just from kernel is < 100. The notify ops's opcode
range is [101-106].
Then we can process all ops consistently.
Cf. func (c *Connection) SetNotifyContext(op interface{})
            (context.Context, error) {

see pr code.

connection.go
inevity: Done.

fuseops/ops.go
inevity: Done.

fishpadak:
Hello. Any progress on this PR?

inevity:
> Hello :D Any progress on this PR?

Dont know why jacobsa not to merge or give some advice.

stapelberg:
Would you mind rebasing this on the current master branch please? We fixed the travis config in the meantime.

inevity:
Just have rebased on master branch.

connection.go
```diff
@@ -497,6 +541,29 @@ func (c *Connection) Reply(ctx context.Context, opErr error) {
 	}
 }
 
+// The NotifyKernel is same as Reply func of Connection.But the diff is
+// that the func only send to kernel.
+func (c *Connection) NotifyKernel(ctx context.Context) {
+
+	// we should get outmsg from context
+	var key interface{} = contextKey
+	foo := ctx.Value(key)
```
stapelberg: Can you explain why we need to pull this value out of the context instead of just passing it as a parameter to NotifyKernel?

inevity: Just do like the reply func.  Of course you can also just pass the value of the context as a parameter to NotifyKernel.

stapelberg: Thanks. The reply mechanism needs to use a context because it passes a value (https://github.com/jacobsa/fuse/blob/659cc51b20ee0bf7bf68875df8ea1a0e0704ba42/connection.go#L406) across API boundaries (to the file system implementation and back).

This is not the case here AIUI, so please change the function signature to not take a context, but just take the opState parameter directly.

inevity: I rethink this case, even the notify func is  diff with the normal op which need reply function. But the context use in notify func  is same as the reply func.   Just the context is created by our func (s *fileSystemServer) Invalidateentry/inode.   And now we call unity all notify op and the normal op.  
For the correct context use, we should alway use the context for notify case. see func (s *fileSystemServer) InvalidateInode for opsInFlight.Done.  

Maybe i am not understandering the issue on detail.

stapelberg: > I rethink this case, even the notify func is diff with the normal op which need reply function. But the context use in notify func is same as the reply func. Just the context is created by our func (s *fileSystemServer) Invalidateentry/inode. And now we call unity all notify op and the normal op.

If I understand correctly, you’re saying having both functions (NotifyKernel and Reply) take arguments via a context is good because they are similar, yes?

While similarity is good, I would argue that in this case, it’s clearer if a context is not used unless absolutely necessary. Since we’re creating a context but then immediately call the function (instead of passing the context somewhere else), I don’t think we need to use a context.

Contexts are used (aside from deadlines and cancellation) to propagate data across API boundaries. In this case, we don’t need to cross API boundaries, so we shouldn’t use a context if we can use a simple variable instead.

inevity:
Can you give some comments about the context commit?
I use context in SetNotifyContext,so this func SetNotifyContext  have nothing to modify, only one bug to fix.

Then change NotifyKernel(opstate opState) as this signature.

And the funcs which  calling notifykernel  in fuseutil/file_system.go Where have some dup code, you can give some detail advice.

inevity:
Wait...

fuseutil/file_system.go
```diff
@@ -123,6 +125,77 @@ func (s *fileSystemServer) ServeOps(c *fuse.Connection) {
...
+	c := s.GetMfs().Conn.(*fuse.Connection)
+	op := &fuseops.NotifyInvalInodeOp{
+		Ino: fuseops.InodeID(ino),
+		Off: off,
+		Len: len,
+	}
+	ctx, _ := c.SetNotifyContext(op)
+	var key interface{} = contextKey
+        foo := ctx.Value(key)
+        opstate, ok := foo.(opState)
```
riking, line 187: The formatting here is a mess.

mount.go
```diff
@@ -27,6 +28,11 @@ type Server interface {
 	// until all operations have been responded to. Must not be called more than
 	// once.
 	ServeOps(*Connection)
+	InvalidateEntry(fuseops.InodeID, string) error
+	InvalidateInode(fuseops.InodeID, int64, int64) error
+	NotifyDelete(fuseops.InodeID, fuseops.InodeID, string) error
+	SetMfs(*MountedFileSystem)
+	GetMfs() *MountedFileSystem
 }
```
riking, line 36: This is a lot of extra methods to be adding to the very simple Server interface, which could plausibly have been proxied in client code..

PR #36, opened 2017-11-27
Hard links implementation

srdjanrilak:
In some situations it's very useful to have hard links in FUSE based file system.
This change will allow the hard links creation.

The change contains very minimal piece of code required to make OpLink operation possible:
- new CreateLinkOp struct is introduced and CreateLink call is added to FileSystem interface,
- code for serialization and deserialization is implemented for the new operation.

### DIFF
--- a/conversions.go
+++ b/conversions.go
@@ -420,6 +420,32 @@ func convertInMessage(
 			Flags:        fusekernel.InitFlags(in.Flags),
 		}
 
+	case fusekernel.OpLink:
+		type input fusekernel.LinkIn
+		in := (*input)(inMsg.Consume(unsafe.Sizeof(input{})))
+		if in == nil {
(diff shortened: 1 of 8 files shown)


jacobsa:
Thanks for this. Could you please add a test to the `samples/` directory, with similar coverage to what already exists for symlinks?

srdjanrilak:
Thanks for the suggestion.

Some test cases are added. There are following changes in the samples/ directory:
- CreateLink operation is implemented in memfs.
- Test cases are added to check: (1) hard link creation, (2) hard link creation over existing file, (3) hard link deletion, (4) reading file content from linked file, (5) parallel creation of hard links.
- Existing negative test case for hard link is removed.

Is this test coverage ok?

jacobsa:
Thanks!

PR #37, opened 2017-11-27
Add recover routine to filesystem

srdjanrilak:
Sometimes panic occurs in code that is in charge of processing a FUSE operation. In can be very useful (for debugging purposes) to be able to recover from that panic in order to (1) log the stack trace (or do whatever is required for debugging the panic) and (2) shut down the file system gracefully (or re-panic).

This change introduces a mechanism to do so.

Alongside NewFileSystemServer function a new, similar, function NewFileSystemServerWithRecover is added. One of the parameters of the new function is a lambda. The lambda will be executed if panic occurs (that is, if recover() return a non-nil result), and whatever is returned by recover() will be set as the lambda's argument. If the lambda in NewFileSystemServerWithRecover is set to nil, everything will work as if NewFileSystemServer is called. ServeOps is changed a bit so defer is called only once (to minimize impact of the change on the performance).

### DIFF
--- a/fuseutil/file_system.go
+++ b/fuseutil/file_system.go
@@ -82,15 +82,59 @@ type FileSystem interface {
 // cf. http://goo.gl/jnkHPO, fuse-devel thread "Fuse guarantees on concurrent
 // requests").
 func NewFileSystemServer(fs FileSystem) fuse.Server {
-	return &fileSystemServer{
-		fs: fs,
+	return NewFileSystemServerWithRecover(fs, nil)
+}
+
+// Create a fuse.Server that handles ops by calling the associated FileSystem
+// method.Respond with the resulting error. Unsupported ops are responded to
+// directly with ENOSYS.
+//
+// If panic occurs in filesystem server, panicHandler will be called with it's
+// argument set to whatever 'recover' returns.
+// The recommended behavior for this function is to handle the exception and
+// to re-panic or stop the execution gracefully.
+// If panicHandler is nil, it will be ignored.
+//
+// Each call to a FileSystem method (except ForgetInode) is made on
+// its own goroutine, and is free to block. ForgetInode may be called
+// synchronously, and should not depend on calls to other methods
+// being received concurrently.
+//
+// (It is safe to naively process ops concurrently because the kernel
+// guarantees to serialize operations that the user expects to happen in order,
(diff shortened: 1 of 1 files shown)


jacobsa:
Can you help me understand what you would want to do with this? What would a typical panic handler look like?

srdjanrilak:
Sure. The typical use case is to supply a function to NewFileSystemServerWithRecover that will log recover message, stack trace or any other information helpful for debugging the panic in a file and then stop the execution. Something like this:

fs := &memFS{
		inodes: make([]*inode, fuseops.RootInodeID+1),
		uid:    uid,
		gid:    gid,
	}

fsServer := fuseutil.NewFileSystemServerWithRecover(fs, func (pnc interface{}){
		if pnc != nil {
				ioutil.WriteFile("/my/log/file", debug.Stack(), 0400)
				os.Exit(1)
		}
	})

This can be helpful when redirecting stderr to a file is not handy, for example.

jacobsa:
I think you can build this already without needing to add explicit support to the API by writing your own implementation of the server interface that wraps the result of NewFileSystemServer, right? It would do a `recover` in its `ServeOps` method and run that code there.

srdjanrilak:
I'm sorry about the late response.

There is a small issue with wrapping fuse.Server to recover from the panics. In this line https://github.com/jacobsa/fuse/blob/c4e473376f7d5be650b11657ded3afb1cd80ad7c/fuseutil/file_system.go#L122
handleOp is started in a separate goroutine.

If we have a ServeOps implementation like following

type wrapper struct {
    next fuse.Server
}

func (w *wrapper) ServeOps(c *fuse.Connection) {
    defer func(){
        rec := recover()
        if rec != nil {
            //...
        }
        }()

    w.next.ServeOps(c)
}

the deferred recover() will return nil (or not be executed) because the panic occurs in a separated goroutine (the one that executes handleOp, not the one that executes chained ServeOps calls).

Did you have different wrapper implementation in mind?

I do realize that proposed change in the API is a bit ugly. If you have a suggestion on how it can be improved, I'd like to hear it.

jacobsa:
Ah yes, good point.

I still do think it's a bit weird to formalize this into an API. Is it crazy to ask you to write a wrapper at the `FileSystem` level? Or even to add your own implementation of `fileSystemServer` that recovers as appropriate?

srdjanrilak:
Those two are possible solutions. Thanks for the input.

Issue #14, opened 2016-01-17
Cannot build - "missing function body"


10098:
This package isn't building on my system.

I am using the following command:  go get -u github.com/jacobsa/fuse

I am getting the following errors:
../../go/src/github.com/jacobsa/fuse/internal/buffer/runtime.go:22: missing function body for "memclr"
../../go/src/github.com/jacobsa/fuse/internal/buffer/runtime.go:27: missing function body for "memmove"

Looking at these files hints at the fact that those are supposed to be entry points for their C counterparts, but I don't know Go well enough to figure out what's really going on.

I am running Arch Linux ARM on a Raspberry Pi 2.

10098:
Since I was building on a Raspberry Pi 2, the assembly file "runtime.s" was being ignored, because its "+build" directive didn't include my target architecture. Removing that line fixed the problem.

jacobsa:
Thanks, can you send a pull request that fixes the issue?

jacobsa:
Closing due to inactivity.

PR #38, opened 2017-12-19
Fix arm build.

DarthGandalf:
I don't know what this list means, and probably it should be removed.
Ref #14

### DIFF
--- a/internal/buffer/runtime_go1.8.s
+++ b/internal/buffer/runtime_go1.8.s
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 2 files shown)


jacobsa:
The list is the set of architectures for which this file is built, and it definitely shouldn't be removed—the file only works for certain architectures.

Are you sure that it works for arm? Did you run the tests? The file from which it was branched had to add [particular support](https://github.com/golang/go/blob/d8c6dac/src/os/signal/sig.s#L11-L13) for arm.

DarthGandalf:
Sorry, you're right, it doesn't work, `go test` hangs. With those 3 lines it still hangs.

Issue #17, opened 2016-08-24
go 1.7 support


allspace:
"golang.org/x/net/context"

this has been builtin package in 1.7.

jacobsa:
It's true, but the two packages are functionally equivalent and the code continues to build under Go 1.7. Have you seen an actual problem?

allspace:
yes, look at the package name "golang.org/x/net/context"

with 1.7 builtin, it should be just "context"

my workaround is to create a directory structure golang.org/x/net, and link context with builtin one.

jacobsa:
Again, the old package continues to exist. As far as I know, it is still compatible with Go 1.7. I certainly haven't seen any trouble on Linux or OS X. Okay yes it requires you to download the package, but `go get -u` makes that easy.

Anyway, please feel free to send a pull request.

kahing:
this is fixed by https://github.com/jacobsa/fuse/pull/39 ?

PR #39, opened 2018-02-07
replace imports of "golang.org/x/net/context" with "context"

saracen:
Go 1.10 should be released this month:
> The go fix tool now replaces imports of "golang.org/x/net/context" with "context". (Forwarding aliases in the former make it completely equivalent to the latter when using Go 1.9 or later.)

This is related to issue https://github.com/jacobsa/fuse/issues/17.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
   include:
     - os: linux
       language: go
-      go: 1.8.1
+      go: 1.9.3
       # Use the virtualized Trusty beta Travis is running in order to get
       # support for installing fuse.
       #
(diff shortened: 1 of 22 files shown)


saracen:
This is currently breaking because https://github.com/jacobsa/syncutil hasn't been updated to use the latest "context" either. I believe if being built with go 1.9, this wouldn't be an issue because net/context is aliased to "context".

I can create a PR for syncutil too, or upgrade to be building on go 1.9, or do both?

jacobsa:
Go 1.9 is fine with me. Thank you!

saracen:
Just a heads up, it looks like the OSX build failed.

PR #40, opened 2018-03-03
Improve Uid & Gid support

manvalls:
The purpose of this PR is to allow file systems to:
1. Create inodes with the UID and GID of the calling process
2. Support chown

### DIFF
--- a/conversions.go
+++ b/conversions.go
@@ -41,7 +41,9 @@ func convertInMessage(
 	inMsg *buffer.InMessage,
 	outMsg *buffer.OutMessage,
 	protocol fusekernel.Protocol) (o interface{}, err error) {
-	switch inMsg.Header().Opcode {
+	header := inMsg.Header()
+
+	switch header.Opcode {
 	case fusekernel.OpLookup:
 		buf := inMsg.ConsumeBytes(inMsg.Len())
 		n := len(buf)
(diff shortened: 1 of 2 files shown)


stapelberg:
Would you mind rebasing this on the current master branch please? We fixed the travis config in the meantime.

conversions.go
```diff
@@ -41,7 +41,9 @@ func convertInMessage(
 	inMsg *buffer.InMessage,
 	outMsg *buffer.OutMessage,
 	protocol fusekernel.Protocol) (o interface{}, err error) {
-	switch inMsg.Header().Opcode {
+	header := inMsg.Header()
+
+	switch header.Opcode {
```
stapelberg: nit: use `switch header := inMsg.Header(); header.Opcode` to scope the `header` variable to the `switch` block only (smaller scopes are generally preferable)

fuseops/ops.go
```diff
@@ -150,6 +150,8 @@ type SetInodeAttributesOp struct {
 	Mode  *os.FileMode
 	Atime *time.Time
 	Mtime *time.Time
+	Uid   *uint32
```
stapelberg: Is there a reason why you’re adding Uid/Gid only to some operations, but not to all? Isn’t the uid/gid field set by the kernel for all requests?

PR #45, opened 2018-09-19
Logger interface

tsavola:
Allows custom logging backends. (It's also more gopheric.)

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"log"
 	"os"
 	"path"
 	"runtime"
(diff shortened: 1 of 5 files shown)


tsavola:
log.Logger is an unnecessary step in the call chain, and accepting an interface is nicer API design than depending on a specific implementation. But since the log.Logger support needs to stay for backward-compatibility, I can understand that this may not seem appealing as it would clutter the API (opposite of the desired effect).

jacobsa:
I don't mind refactoring, but I fail to see what this interface gives you that log.New doesn't. Does this open up some possibility for you that wasn't there before? If so, what?

tsavola:
As you pointed out, since log.Logger can be used with a custom io.Writer, it can be "re-generalized" if necessary. I'm wrapping fuse in another package, and I didn't want to perpetuate the specific logger type requirement in its API.

I don't have a concrete requirement for using anything other than log.Logger with fuse at this point.

PR #46, opened 2018-10-14
Split fuse mount and server launch for flexible use

lkpdn:
When we want a file system to work well with automounting, we may want more flexibility in `fuse.Mount`, i.e. mounting fuse file system in the original process and serving requests in daemonized process(es).
This patch does it in a completely backward-compatible way.

For reference, based on this patch, I cooked up a patch to make `kahing/goofys.git` work well with automounting. (https://github.com/kahing/goofys/pull/369)

Please review, thanks.

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -29,41 +29,18 @@ type Server interface {
 	ServeOps(*Connection)
 }
 
-// Mount attempts to mount a file system on the given directory, using the
-// supplied Server to serve connection requests. It blocks until the file
-// system is successfully mounted.
-func Mount(
-	dir string,
+// FuseServe attempts to start the supplied fuse server with the file discriptor
+// to poll against. It blocks until the file system is successfully mounted.
+func FuseServe(
 	server Server,
-	config *MountConfig) (mfs *MountedFileSystem, err error) {
-	// Sanity check: make sure the mount point exists and is a directory. This
-	// saves us from some confusing errors later on OS X.
-	fi, err := os.Stat(dir)
-	switch {
-	case os.IsNotExist(err):
-		return
-
-	case err != nil:
-		err = fmt.Errorf("Statting mount point: %v", err)
-		return
-
-	case !fi.IsDir():
-		err = fmt.Errorf("Mount point %s is not a directory", dir)
-		return
-	}
-
+	dir string,
+	devFd uintptr,
+	config *MountConfig,
+	ready chan error) (mfs *MountedFileSystem, err error) {
 	// Initialize the struct.
 	mfs = &MountedFileSystem{
 		dir:                 dir,
 		joinStatusAvailable: make(chan struct{}),
-	}
-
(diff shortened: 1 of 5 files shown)


mount.go
```diff
@@ -96,6 +76,68 @@ func Mount(
...
+	// Begin the mounting process, which will continue in the background.
+	dev, err = mount(dir, config, ready)
+	if err != nil {
+		err = fmt.Errorf("mount: %v", err)
+		return
+	}
+	return
+}
+
+func IsMounted(
```
jacobsa: Why do you need this? Please also add documentation.

mount.go
```diff
@@ -96,6 +76,68 @@ func Mount(
...
 	}
+	return
+}
+
+// FuseMount attempts to mount a file system on the given directory. It can block
+// fuse server setup completion using supplied ready channel.
+func FuseMount(
+	dir string,
+	config *MountConfig,
+	ready chan error) (dev uintptr, err error) {
```
jacobsa: These use of this channel isn't clear to me. Please elaborate in the documentation what it's for, and when an error may be delivered on the channel versus returned from the function. Please also mark it as a send-only channel to make the intention clear.

mount.go
```diff
@@ -29,49 +29,29 @@ type Server interface {
...
 }
 
-// Mount attempts to mount a file system on the given directory, using the
-// supplied Server to serve connection requests. It blocks until the file
-// system is successfully mounted.
-func Mount(
-	dir string,
+// FuseServe attempts to start the supplied fuse server with the file discriptor
+// to poll against. It blocks until the file system is successfully mounted.
+func FuseServe(
```
jacobsa: Could you please describe in the documentation for this and the function below when you would want to use them versus just using Mount? Having three entry points is confusing; we should at least have the more complex ones give a pointer to the simpler ones.

mount.go
lkpdn: This is just a util func and some still do not need this. Now I'm wondering this should be implemented in client side, not in this jacobsa/fuse.git. Perhaps I'll remove these.

mount.go
lkpdn: Okay I'll add a comment above the func definition, explaining what it's for + when error delivered.

> Please also mark it as a send-only channel to make the intention clear.

I'll push a commit later. Thanks.

mount.go
lkpdn: I'll add a comment. Thanks.

lkpdn:
Thanks for your quick response.

> Hi, could you please give some more info on what this is for? For example, could you summarize how you intend to use this? I read the goofyfs pull request description, but it doesn't make any sense to me (perhaps because I'm not familiar with goofyfs and automount). Some intermediate level of detail would be helpful.

Could you please wait for a while until sufficient discussion having been done on the goofys PR? Then I'll mention you here on this PR for notification.

PR #58, opened 2018-12-26
convertAttributes: set S_ISUD bit from os.FileMode’s os.ModeSetuid

stapelberg:
Before this change, the package effectively silently stripped the suid bit.


stapelberg:
The macOS travis failure seems unrelated to me.

PR #59, opened 2018-12-26
try mounting without fusermount(1) first

stapelberg:
This requires root privileges or the CAP_SYS_ADMIN capability, but allows supporting -o suid.

I also addressed a bunch of readability-related nits while touching the code.

### DIFF
--- a/mount_linux.go
+++ b/mount_linux.go
@@ -2,29 +2,21 @@ package fuse
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"net"
 	"os"
 	"os/exec"
 	"syscall"
+
+	"golang.org/x/sys/unix"
 )
 
-// Begin the process of mounting at the given directory, returning a connection
-// to the kernel. Mounting continues in the background, and is complete when an
-// error is written to the supplied channel. The file system may need to
-// service the connection in order for mounting to complete.
-func mount(
-	dir string,
-	cfg *MountConfig,
-	ready chan<- error) (dev *os.File, err error) {
-	// On linux, mounting is never delayed.
-	ready <- nil
-
+func fusermount(dir string, cfg *MountConfig) (*os.File, error) {
 	// Create a socket pair.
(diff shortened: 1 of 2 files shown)


mount_linux.go
```diff
@@ -126,10 +126,11 @@ var mountflagopts = map[string]func(uintptr) uintptr{
 var errFallback = errors.New("sentinel: fallback to fusermount(1)")
 
 func directmount(dir string, cfg *MountConfig) (*os.File, error) {
-	dev, err := os.OpenFile("/dev/fuse", os.O_RDWR, 0644)
+	fd, err := syscall.Open("/dev/fuse", syscall.O_RDWR, 0644)
```
hugelgupf: I'd also recommend a comment here saying why this isn't an os.OpenFile call

stapelberg: Done.

kungf:
hi, @stapelberg  can you help me,with this patch when i mount fuse with root, hit the err
"mount: invalid argument"
seems mount source not work, can you give some advise?

but normal user is ok!

stapelberg:
Can you open an issue for that please? In the issue, can you supply an strace log (use strace -fvy -s2048 -o /tmp/strace.log in front of your command), and can you let us know how to reproduce the problem?

kungf:
@stapelberg 

> Can you open an issue for that please? In the issue, can you supply an strace log (use strace -fvy -s2048 -o /tmp/strace.log in front of your command), and can you let us know how to reproduce the problem?

https://github.com/jacobsa/fuse/issues/76

thank you very much!

kungf:
@stapelberg hi， can you tell something about why root not use fusermount？

stapelberg:
This is a requirement for supporting the `-o suid` option. Also, it helps in early-boot environments where fusermount might not be available.

kungf:
> This is a requirement for supporting the `-o suid` option. Also, it helps in early-boot environments where fusermount might not be available.

got it, thank you

PR #63, opened 2019-05-16
Add open flags to OpenFileOp

rminnich:
The open flag indicates the type of access we need.

Add a test to the memfs server to make sure an open
is blocked if the flags and permissions don't match.

Signed-off-by: Ronald G. Minnich <rminnich@gmail.com>


fuseops/ops.go
```diff
@@ -593,6 +593,9 @@ type OpenFileOp struct {
 	// layer. This allows for filesystems whose file sizes are not known in
 	// advance, for example, because contents are generated on the fly.
 	UseDirectIO bool
+
+	// Open flags. See os.OpenFile
```
jacobsa:
```suggestion
	// User-provided flags to open(2). See http://go/godoc/os/#OpenFile.
```

samples/memfs/memfs.go
```diff
@@ -615,6 +616,24 @@ func (fs *memFS) OpenFile(
 	if !inode.isFile() {
 		panic("Found non-file.")
 	}
+	// What perms do we need?
```
jacobsa: I had trouble understanding what you were doing here at first. Could you make
the code read something like:


    // Check that the open mode is legal according to the file
    // permissions.
    var permissionBits int
    switch (...) {
        ...
    }
    
    if (...) {
      err = EACCESS
      return
    }

samples/memfs/memfs.go
```diff
@@ -615,6 +616,24 @@ func (fs *memFS) OpenFile(
...
+	var perm int
+	switch op.Flag & unix.O_ACCMODE {
+	case unix.O_RDONLY:
+		perm = 0400
+	case unix.O_WRONLY:
+		perm = 0200
+	case unix.O_RDWR:
+		perm = 0600
+	// This is legal, at least on linux
+	case unix.O_ACCMODE:
```
jacobsa: This case is weird, and I think it's just an artifact of using a switch
statement, right? Why not something like:

    if flags & unix.O_RDONLY && !(inode.attrs.Mode()&unix.O_RDONLY) {
      return EACCESS
    }
    
    // Ditto with O_WRONLY and O_RDWR
    ...

rminnich: we should not check the attrs ourself. That's the kernel's job. I can change it to the if but I found the case more readable, and was deliberately including a case that seems weird but works: flags is unix.O_ACCMODE. But I would strongly argue that we let the kernel do what it does w.r..t EACCES

samples/memfs/memfs.go
rminnich: we should not be checking attributes, since they can change behind our back.

samples/memfs/memfs.go
rminnich: Also, this is another original defect of unix: rdwr is 2, wronly 1, rdonly 0, so it's not actually a bit mask. Hence the case. The real test should be on a value comparison, not a bitmask. A shame, but it's what they did.

rminnich: oh yeah, and that case is very deliberate: I tested it on linux before I added it. Again, I broke it out b/c it's not obvious.

rminnich: oops, I was thinking about the fs I am writing, not memfs, sorry. nvm.

rminnich: but if you want it to work your proposed way that's fine too.

jacobsa: Sorry, I'm totally lost at this point. All I want you to do in `memfs_test` is somehow confirm that the `open(2)` flags correctly come through to the file system. Maybe you are attempting to do that here, but I don't understand how. Could you please improve the commenting, here or in the test, to explain how you're confirming this?

PR #68, opened 2019-10-21
fix: directIO read size can exceeds buffer size

shuoranliu:
Since the package uses a fixed-size memory pool for out messages,
it is important to tell fuse kernel module what the maximum read size is.
Otherwise, a large read will cause a panic.

Right now, max readahead size can only affect buffered read. The kernel
might still send a large directIO read in spite of the max readahead size.
We need to specify "max_read" option during mount.

Signed-off-by: Shuoran Liu <shuoranliu@gmail.com>


mount_linux.go
```diff
@@ -29,6 +32,8 @@ func fusermount(dir string, cfg *MountConfig) (*os.File, error) {
 	// Start fusermount, passing it a buffer in which to write stderr.
 	var stderr bytes.Buffer
 
+	cfg.Options["max_read"] = strconv.Itoa(buffer.MaxReadSize)
```
stapelberg: I think you’ll need to ensure cfg.Options is non-nil here, see the travis test failure.

shuoranliu: Thanks, I've made a modification. To reproduce the issue, just try to invoke a read system call with the size of 512Kbytes which is larger than MaxReadSize, and with O_DIRECT flag.

stapelberg: Thanks for fixing this! Would it be possible to add a unit test, too, which exercises this behavior? It should fail before your fix commit, and pass afterwards, so that we don’t accidentally re-introduce this bug at some point down the road.

shuoranliu: Is there a guide on how to write ci test cases? I believe it is better to write a test case from user space file system. BTW, I have no idea why ci failed again.

stapelberg: Not sure I understand your question—there are no “ci test cases”, the CI just runs the go tests of this project :)

This particular failure might have been a flake (sometimes the CI is overloaded), so I have restarted the run.

PR #69, opened 2019-10-24
Don't log ERANGE for ListXattrOp.

lukealonso:
ListXattrOp behaves similarly to GetXattrOp in that you have to return ERANGE in order to indicate the size of the attribute list, so logging errors isn't warranted.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -432,7 +432,10 @@ func (c *Connection) shouldLogError(
 		if err == syscall.ENOENT {
 			return false
 		}
-
+	case *fuseops.ListXattrOp:
+		if err == syscall.ENODATA || err == syscall.ERANGE {
+			return false
+		}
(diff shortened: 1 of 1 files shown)


connection.go
```diff
@@ -432,7 +432,10 @@ func (c *Connection) shouldLogError(
 		if err == syscall.ENOENT {
 			return false
 		}
-
+	case *fuseops.ListXattrOp:
```
stapelberg: I think you can just add this type to the case statement below, like so:

```
case *fuseops.GetXattrOp, *fuseops.ListXattrOp:
```

riking:
@stapelberg - merge and fixup yourself?

stapelberg:
Merged with commit https://github.com/jacobsa/fuse/commit/05606cde59ac2f1595bdd9bfc44385182e68bb45

PR #72, opened 2019-12-10
fix getxattr and listxattr

kahing:
previously, this will fail if /mnt/file doesn't have an xattr:

```
listxattr("/mnt/file", 0x7fe8b3686830, 256) = -1 EIO (Input/output error)
```

We should be returning the actual size only if the input size is
zero. Related issue is if the filesystem returns ERANGE, we should
propagate that error instead of returning the actual size.

Replaced go-xattr usage with x/sys/unix so we can test this.

### DIFF
--- a/conversions.go
+++ b/conversions.go
@@ -603,17 +603,6 @@ func (c *Connection) kernelResponse(
 	if opErr != nil {
 		handled := false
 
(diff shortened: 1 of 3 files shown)


stapelberg:
Thanks!

PR #85, opened 2020-06-22
Add ARM support.

Bren2010:
This PR adds 32-bit ARM support. Tested on a Raspberry Pi (and verified that it actually works).


stapelberg:
Thank you!

PR #87, opened 2020-06-30
Set max file/folder name length.

Bren2010:
Currently, trying to create a file/folder with a name longer than 1 character fails. This fixes that by setting the max name length in StatFS.

### DIFF
--- a/conversions.go
+++ b/conversions.go
@@ -706,6 +706,7 @@ func (c *Connection) kernelResponseForOp(
 		out.St.Bavail = o.BlocksAvailable
 		out.St.Files = o.Inodes
 		out.St.Ffree = o.InodesFree
+		out.St.Namelen = 255
 
 		// The posix spec for sys/statvfs.h (http://goo.gl/LktgrF) defines the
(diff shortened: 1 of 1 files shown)


stapelberg:
> Currently, trying to create a file/folder with a name longer than 1 character fails

On which operating system etc.? I’m surprised to read this.

Why are you setting the name length to 255 unconditionally? Shouldn’t it be set to the actual length?

Bren2010:
On Ubuntu, using the Nautilus helper. This is setting the max name length, where 255 is [normal](https://github.com/torvalds/linux/blob/851ca779d110f694b5d078bc4af06d3ad37169e8/include/uapi/linux/limits.h#L12) I think

Screenshot from 2020-07-01 08-23-30

riking:
Can you also add Namelen to the configuration struct, and default the value to 255 if the config is 0?

stapelberg:
Ah, now I understand: this fills in the `f_namelen` field in `statfs(2)` syscalls, which programs use to figure out how long to make their file names.

PR #88, opened 2020-07-05
posix_test: Reenable O_APPEND tests by using raw syscall

riking:
This re-enables the tests disabled in commit fbd48a3.


stapelberg:
Thanks!

PR #92, opened 2020-10-15
Add pid in all fuse ops

xinmeigui-db:
Add pid in all fuse ops

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -409,7 +409,8 @@ func (c *Connection) ReadOp() (_ context.Context, op interface{}, _ error) {
 			debugMsg := describeRequest(op)
 			c.debugLog(inMsg.Header().Unique, 1, "<- %s", debugMsg)
 			if strings.Contains(debugMsg, "NO PID FOR OP") {
-				panic("wow wow wow")
+				// Added for debugging temporarily.
+				panic("==NO PID FOR OP==")
 			}
 		}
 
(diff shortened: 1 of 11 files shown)


xinmeigui-db:
Hi @stapelberg , would appreciate some comments & feedback!

stapelberg:
Adding the caller process id to all ops sounds okay to me in general, but:

1.  Why do you need to disable writeback caching for the pid to become available?
2. Your pull request seems to contain a lot of unrelated commits? Can you clean it up please? (rebase)

Thanks,

xinmeigui-db:
> Adding the caller process id to all ops sounds okay to me in general, but:
> 
> 1. Why do you need to disable writeback caching for the pid to become available?
> 2. Your pull request seems to contain a lot of unrelated commits? Can you clean it up please? (rebase)
> 
> Thanks,

Thanks for the review!

1. We disabled the writeback cache because when the write comes from the writeback cache, there is no pid associated with that (the test may be flaky as a result).
2. Rebased!

dotslash:
> Your pull request seems to contain a lot of unrelated commits? Can you clean it up please? (rebase)
If the PR is squash-merged, then this is not relevant. right?

stapelberg:
> If the PR is squash-merged, then this is not relevant. right?

Cleanly separated commits also make the review easier, no matter if they end up in the git history or not :)

Issue #93, opened 2020-12-15
Shift enforced default permissions to an option.


Arcayr:
Hey there,

I'm a user of Goofys (https://github.com/kahing/goofys) which uses this library for pretty much all of its functionality. I moved to it from s3fs-fuse, which is a C application which fulfills more or less the same purpose but at a severely reduced speed. Through testing with a Nextcloud installation I noticed that this library enforces the use of FUSE's `default_permissions` mount option:

https://github.com/jacobsa/fuse/blob/36e01f1b3aaafad29231bd7286dac6c384110502/mount_config.go#L182

I ended up writing a small patch file for Goofys to remove it prior to compiling, but I'd rather it be in upstream and offered as a configuration option, since the current behaviour is that which users of the library have come to expect.

I've got [a branch right now](https://github.com/Arcayr/fuse/tree/no-default-permissions) with a single config flag in `MountOptions` to disable this if required. Is that something you'd be interested in having merged in? It would fix this issue for people shifting across from `s3fs-fuse` in Goofys. I can open a PR if that seems like something you'd be willing to accept.

stapelberg:
Yes, we can make this optional I think. Please send a PR :)

Arcayr:
Fixed in #94.

PR #94, opened 2020-12-16
Switch using Linux "default_permissions" to an optional parameter.

Arcayr:
Discussion from #93.


Arcayr:
The CI build failed but judging by its logs its due to an incorrect CI configuration (it's waiting for output that never arrives).

stapelberg (approved):

PR #95, opened 2021-03-20
Added samples/roloopbackfs

ReallyLiri:
fs for readonly access to an existing physical dir (loopback)

### DIFF
--- a/samples/mount_roloopbackfs/mount.go
+++ b/samples/mount_roloopbackfs/mount.go
@@ -1,0 +1,74 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"github.com/jacobsa/fuse/samples/roloopbackfs"
+	"log"
+	"os"
+
+	"github.com/jacobsa/fuse"
+)
+
+var fPhysicalPath = flag.String("path", "", "Physical path to loopback.")
+var fMountPoint = flag.String("mount_point", "", "Path to mount point.")
+
+var fDebug = flag.Bool("debug", false, "Enable debug logging.")
+
+func main() {
+	flag.Parse()
+
+	debugLogger := log.New(os.Stdout, "fuse: ", 0)
+	errorLogger := log.New(os.Stderr, "fuse: ", 0)
+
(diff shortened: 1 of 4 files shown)


samples/mount_roloopbackfs/mount.go
```diff
@@ -0,0 +1,74 @@
...
+	if *fPhysicalPath == "" {
+		log.Fatalf("You must set --path.")
+	}
+
+	server, err := roloopbackfs.NewReadonlyLoopbackServer(*fPhysicalPath, errorLogger)
+	if err != nil {
+		log.Fatalf("makeFS: %v", err)
+	}
+
+	if *fMountPoint == "" {
```
stapelberg, line 47: Can you check all flags at the beginning? That way, invocations that won’t work fail as early as possible.

samples/mount_roloopbackfs/mount.go
```diff
@@ -0,0 +1,74 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"github.com/jacobsa/fuse/samples/roloopbackfs"
```
stapelberg, line 20: Here and in the other files, can you please separate standard library imports from external imports? In this case, move the `github.com/jacobsa/fuse/samples/roloopbackfs` line after line 24, so that you have 2 groups of imports separated by a blank line.

samples/mount_roloopbackfs/mount.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		log.Fatalf("makeFS: %v", err)
+	}
+
+	if *fMountPoint == "" {
+		log.Fatalf("You must set --mount_point.")
+	}
+
+	_, err = os.Stat(*fMountPoint)
+	if os.IsNotExist(err) {
```
stapelberg, line 52: Stylistically, these two lines can be combined into one:

```go
if _, err := os.Stat(*fMountPoint); os.IsNotExist(err) {
```

But you don’t actually need to check when using `os.MkdirAll`, you can just call `os.MkdirAll` unconditionally.

samples/mount_roloopbackfs/mount.go
```diff
@@ -0,0 +1,74 @@
...
+		os.MkdirAll(*fMountPoint, 0777)
```
stapelberg, line 53: check the error returned here

samples/roloopbackfs/roloopbackfs.go
```diff
@@ -0,0 +1,201 @@
...
+	logger       *log.Logger
+}
+
+var _ fuseutil.FileSystem = &readonlyLoopbackFs{}
+
+// Create a file system that mirrors an existing physical path, in a readonly mode
+
+func NewReadonlyLoopbackServer(loopbackPath string, logger *log.Logger) (server fuse.Server, err error) {
+	_, err = os.Stat(loopbackPath)
+	if err != nil {
```
stapelberg, line 40:
```go
if _, err := os.Stat(loopbackPath); err != nil {
```

stapelberg (commented):
Thanks for your PR!

This looks good overall, left a couple of stylistic comments.

stapelberg (approved):

PR #96, opened 2021-05-21
Add support for max_pages and increase to 256

kpjensen:
This PR lacks many things, but is a start to add support for `max_pages`, which was added in kernel 4.20, fuse version 7.28.

Default values for the new fields in `init_out` were taken from where I think the kernel sets their default values. 
https://github.com/torvalds/linux/blob/9f67672a817ec046f7554a885f0fe0d60e1bf99f/include/uapi/linux/fuse.h#L740-L752 

`max_background` https://github.com/torvalds/linux/blob/27787ba3fa4904422b3928b898d1bd3d74d98bea/fs/fuse/inode.c#L58
`congestion_threshold` https://github.com/torvalds/linux/blob/27787ba3fa4904422b3928b898d1bd3d74d98bea/fs/fuse/inode.c#L61
`time_gran` https://github.com/torvalds/linux/blob/27787ba3fa4904422b3928b898d1bd3d74d98bea/fs/fuse/inode.c#L1263

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,1 +1,0 @@
-module github.com/jacobsa/fuse
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,1 @@
+module github.com/jacobsa/fuse
--- a/internal/fusekernel/fuse_kernel.go
+++ b/internal/fusekernel/fuse_kernel.go
@@ -725,7 +725,8 @@ type InitOut struct {
 	MaxWrite            uint32
 	TimeGran            uint32
 	MaxPages            uint16
-	Unused              [8]uint16
+	MapAlignment        uint16
+	Unused              [8]uint32
 }
 
 type InterruptIn struct {
--- a/internal/fusekernel/fuse_kernel.go
+++ b/internal/fusekernel/fuse_kernel.go
@@ -44,7 +44,7 @@ import (
 // The FUSE version implemented by the package.
 const (
 	ProtoVersionMinMajor = 7
-	ProtoVersionMinMinor = 28
+	ProtoVersionMinMinor = 19
 	ProtoVersionMaxMajor = 7
 	ProtoVersionMaxMinor = 31
 )
--- a/internal/fusekernel/fuse_kernel.go
+++ b/internal/fusekernel/fuse_kernel.go
@@ -724,7 +724,8 @@ type InitOut struct {
 	CongestionThreshold uint16
 	MaxWrite            uint32
 	TimeGran            uint32
-	MaxPages            uint32
+	MaxPages            uint16
+	Unused              [8]uint16
 }
 
 type InterruptIn struct {
--- a/internal/fusekernel/fuse_kernel.go
+++ b/internal/fusekernel/fuse_kernel.go
@@ -44,9 +44,9 @@ import (
 // The FUSE version implemented by the package.
 const (
 	ProtoVersionMinMajor = 7
-	ProtoVersionMinMinor = 8
+	ProtoVersionMinMinor = 28
 	ProtoVersionMaxMajor = 7
-	ProtoVersionMaxMinor = 12
+	ProtoVersionMaxMinor = 31
 )
 
 const (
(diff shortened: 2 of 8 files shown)


stapelberg:
Thanks for the PR!

> This PR lacks many things

So, is it ready to review or no? What’s lacking?

Can you also describe what effect your changes have?

kpjensen:
I think it's ready for review now. What I'm unsure of is backwards compatibility with fuse versions < 7.28. I lowered the minimum minor version to 19 as that's what osxfuse supports. https://github.com/osxfuse/fuse/blob/799455e5af70d486d43516676422a01b156be3d8/include/fuse_kernel.h#L118-L122. 

The effect of these changes is to increase the maximum read and write size on linux from 32 pages to 256 by passing the `max_pages` flags and parameter in the init reply to the kernel. 
This helps increase throughput on the networked filesystem I'm working on https://github.com/dnanexus/dxfuse.

internal/fusekernel/fuse_kernel.go
```diff
@@ -714,12 +716,15 @@ type InitIn struct {
...
-	MaxWrite     uint32
+	Major               uint32
+	Minor               uint32
+	MaxReadahead        uint32
+	Flags               uint32
+	MaxBackground       uint16
+	CongestionThreshold uint16
+	MaxWrite            uint32
+	TimeGran            uint32
+	MaxPages            uint32
```
stapelberg, line 727: This should be `MaxPages uint16` as per https://github.com/torvalds/linux/blame/master/include/uapi/linux/fuse.h, no?

I think we need (or should add nevertheless) the `unused` field as well, no?

go.mod
```diff
@@ -0,0 +1 @@
+module github.com/jacobsa/fuse
```
stapelberg, line 1: I have declared a go module separately in commit https://github.com/jacobsa/fuse/commit/9c4c16288f886ca7266061ed859bcef7b26c22a8, so please remove this addition from your commit :)

stapelberg (changes requested):

stapelberg:
> What I'm unsure of is backwards compatibility with fuse versions < 7.28

What exactly are you unsure about?

go.mod
kpjensen: Removed, thank you.

internal/fusekernel/fuse_kernel.go
kpjensen: Fixed, thank you.

internal/fusekernel/fuse_kernel.go
```diff
@@ -714,12 +716,16 @@ type InitIn struct {
...
-	MaxWrite     uint32
+	Major               uint32
+	Minor               uint32
+	MaxReadahead        uint32
+	Flags               uint32
+	MaxBackground       uint16
+	CongestionThreshold uint16
+	MaxWrite            uint32
+	TimeGran            uint32
+	MaxPages            uint16
```
stapelberg, line 727: After MaxPages, it looks like you’re missing `MapAlignment uint16`? https://github.com/torvalds/linux/blame/c500bee1c5b2f1d59b1081ac879d73268ab0ff17/include/uapi/linux/fuse.h#L754

internal/fusekernel/fuse_kernel.go
```diff
@@ -714,12 +716,16 @@ type InitIn struct {
...
+	Unused              [8]uint16
```
stapelberg, line 728: Unused should be `[8]uint32` instead of `[8]uint16`, no?

stapelberg (commented):
Looks good overall and works in my tests! Just two small comments about the data structure definition

internal/fusekernel/fuse_kernel.go
kpjensen: Fixed, thanks.

internal/fusekernel/fuse_kernel.go
kpjensen: Fixed, thanks.

stapelberg (approved):

PR #104, opened 2021-08-26
Vectored / "zero" (almost zero) copy read

vitalif:
Hi!
Now I have a more complex patch.
Currently FUSE filesystems that cache data in memory always have to copy it into the library-provided read buffer. But in fact it's not really required because there's the 'writev' system call which allows to send data from multiple buffers at once.
So in this patch I reworked the ReadFile operation to take a list of slices as a reply and send them to FUSE via the writev system call. This allows file systems that already have the required data in memory to send it without copying, thus significantly reducing overhead.
The change is incompatible with previous versions of the library, so I want to ask you what you think about it :-) if it's unacceptable to break backwards compatibility then I think it may be possible to implement vectored read as a separate operation and select the implementation the FS should use at the moment of initialisation.

I also did a similar thing for writes, but I'll try to submit it in a separate PR. In fact it'a bit tricky because all writes are received from FUSE in fixed-size buffers and it can't be changed because FUSE doesn't allow to read the request with multiple read() calls, so if a write is smaller than the whole buffer it's actually better to copy it because otherwise
a) a lot of memory will be wasted
b) a new buffer will be allocated and zeroed later, and it takes more time to zero 1 MB than to copy 4 KB of data
The upside is that the "zero-copy" write patch is backwards compatible, as opposed to this one. :)

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -338,7 +338,7 @@ func (c *Connection) readMessage() (*buffer.InMessage, error) {
 
 	// Loop past transient errors.
 	for {
-		// Attempt a reaed.
+		// Attempt a read.
 		err := m.Init(c.dev)
 
 		// Special cases:
@@ -405,7 +405,7 @@ func (c *Connection) ReadOp() (_ context.Context, op interface{}, _ error) {
 
 		// Convert the message to an op.
 		outMsg := c.getOutMessage()
-		op, err = convertInMessage(inMsg, outMsg, c.protocol)
+		op, err = convertInMessage(&c.cfg, inMsg, outMsg, c.protocol)
 		if err != nil {
 			c.putOutMessage(outMsg)
 			return nil, nil, fmt.Errorf("convertInMessage: %v", err)
@@ -510,10 +510,16 @@ func (c *Connection) Reply(ctx context.Context, opErr error) {
 	noResponse := c.kernelResponse(outMsg, inMsg.Header().Unique, op, opErr)
 
 	if !noResponse {
-		err := c.writeMessage(outMsg.Bytes())
-		if err != nil && c.errorLogger != nil {
-			c.errorLogger.Printf("writeMessage: %v %v", err, outMsg.Bytes())
+		var err error
+		if outMsg.Sglist != nil {
+			_, err = writev(int(c.dev.Fd()), outMsg.Sglist)
+		} else {
+			err = c.writeMessage(outMsg.OutHeaderBytes())
 		}
+		if err != nil && c.errorLogger != nil {
+			c.errorLogger.Printf("writeMessage: %v %v", err, outMsg.OutHeaderBytes())
+		}
+		outMsg.Sglist = nil
 	}
 }
 
--- a/fuseops/ops.go
+++ b/fuseops/ops.go
@@ -654,6 +654,37 @@ type ReadFileOp struct {
 	OpContext OpContext
 }
 
+// Vectored read - same as ReadFileOp, but the buffer isn't provided by the library.
+// The file system returns a list of slices instead.
+type VectoredReadOp struct {
+	// The file inode that we are reading, and the handle previously returned by
+	// CreateFile or OpenFile when opening that inode.
+	Inode  InodeID
+	Handle HandleID
+
+	// The offset within the file at which to read.
+	Offset int64
+
+	// The size of the read.
+	Size int64
+
+	// Set by the file system: data to send back to the client.
+	Data [][]byte
+
+	// Set by the file system: the number of bytes read.
+	//
+	// The FUSE documentation requires that exactly the requested number of bytes
+	// be returned, except in the case of EOF or error (http://goo.gl/ZgfBkF).
+	// This appears to be because it uses file mmapping machinery
+	// (http://goo.gl/SGxnaN) to read a page at a time. It appears to understand
+	// where EOF is by checking the inode size (http://goo.gl/0BkqKD), returned
+	// by a previous call to LookUpInode, GetInodeAttributes, etc.
+	//
+	// If direct IO is enabled, semantics should match those of read(2).
+	BytesRead int
+	OpContext OpContext
+}
+
 // Write data to a file previously opened with CreateFile or OpenFile.
 //
 // When the user writes data using write(2), the write goes into the page
--- a/fuseutil/not_implemented_file_system.go
+++ b/fuseutil/not_implemented_file_system.go
@@ -138,6 +138,12 @@ func (fs *NotImplementedFileSystem) ReadFile(
 	return fuse.ENOSYS
 }
 
+func (fs *NotImplementedFileSystem) VectoredRead(
+	ctx context.Context,
+	op *fuseops.VectoredReadOp) error {
+	return fuse.ENOSYS
+}
+
 func (fs *NotImplementedFileSystem) WriteFile(
 	ctx context.Context,
 	op *fuseops.WriteFileOp) error {
--- a/internal/buffer/out_message.go
+++ b/internal/buffer/out_message.go
@@ -16,7 +16,6 @@ package buffer
 
 import (
 	"fmt"
-	"log"
 	"reflect"
 	"unsafe"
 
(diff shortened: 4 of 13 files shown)


stapelberg:
Thanks for your PR!

I haven’t reviewed the code in detail yet (it’s pretty intricate stuff and I’m not very familiar with this part of the code base yet…), but what I can say is that we need a solution that does not break backwards compatibility, as this package has quite a number of users.

Also, can you share measurements before and after? How much faster do reads get with your approach?

vitalif:
OK, I'll try to rework it into a separate "VectoredReadOp" or something like that.
Regarding benchmarks:
I've just made an example which emulates a 1 TB file backed by 1 MB of random bytes which are repeated over and over again in the file: `readbenchfs` - https://github.com/vitalif/fusego/blob/zerocopy-read/samples/mount_readbenchfs/mount.go
In my case without zerocopy it serves reads at 633 MB/s, and with zerocopy at 1.7 GB/s.
You can try it with `dd if=mount/ya bs=1M iflag=direct status=progress` or `fio -name=test -ioengine=psync -direct=1 -bs=1M -rw=read -filename=mount/test`. First checkout the commit before vectored
In a real-world example (yandex-cloud/geesefs that I'm working on) vectored read improves sequential read speed from ~290 MB/s to ~470 MB/s (varying from run to run of course). Goofys was even faster in this test (520 MB/s) but that's because it doesn't suffer from the additional copy because it doesn't buffer the data in memory.

vitalif:
OK, I reworked the patch - now it's backwards compatible. Please review :)

It seems that zero-copy was not the only one improvement, because usual "non-zero-copy" reads also improve with it.

My results:
- 1 core i.e. `GOMAXPROCS=1 ./readbenchfs`
  - before patch: 390 MB/s
  - non-vectored read after patch: 830 MB/s
  - vectored read: 1200 MB/s
- all cores (I have 16 of them but only a few were utilized)
  - before patch: 900 MB/s
  - non-vectored read after patch: 1700 MB/s
  - vectored read: 2200 MB/s

vitalif:
On my laptop the improvement is even more dramatic, probably because of the newer kernel 5.10 with 1M maxread/maxwrite (by the way, 1M maxread/maxwrite seemingly doesn't work without O_DIRECT).
`GOMAXPROCS=1 ./readbenchfs --mount_point dir` then `dd if=dir/test of=/dev/null bs=1M status=progress iflag=direct` = 500 MB/s
`GOMAXPROCS=1 ./readbenchfs --mount_point dir --vectored` then `dd if=dir/test of=/dev/null bs=1M status=progress iflag=direct` = 2.8 GB/s

fuseops/ops.go
```diff
@@ -654,6 +654,37 @@ type ReadFileOp struct {
...
+	Handle HandleID
+
+	// The offset within the file at which to read.
+	Offset int64
+
+	// The size of the read.
+	Size int64
+
+	// Set by the file system: data to send back to the client.
+	Data [][]byte
```
stapelberg, line 672: I wonder if we could add a new field (e.g. VectoredData, and Size) to the existing ReadFileOp type?

If UseVectoredRead is enabled, VectoredData is used, otherwise Dst is used. Size can be set in either case.

stapelberg (commented):
Thanks for the benchmark! Sounds like an exciting performance improvement :)

This code is pretty low-level, so I’ll need to do another read and understand how the code worked before/after.

In the meantime, see my comment regarding using the same file operation type please.

fuseops/ops.go
vitalif: Maybe, in fact that came first to my mind, but then I decided that a separate operation was more explicit, i.e. you get a ReadFileOp and you're 100% sure that it doesn't have empty .Dst. Also if both variants are merged in the same op then Size should probably be set for both variants and someone may begin using that field in the future... But that's up to you, I think I can merge them back if you insist :)

vitalif:
Hi, did you have time to look into the code? :-)

vitalif:
@stapelberg ? :-)

internal/buffer/out_message.go
```diff
@@ -100,51 +74,63 @@ func (m *OutMessage) ShrinkTo(n int) {
...
 	return
 }
 
 // Len returns the current size of the message, including the leading header.
 func (m *OutMessage) Len() int {
-	return OutMessageHeaderSize + m.payloadOffset
+	if m.Sglist == nil {
+		return OutMessageHeaderSize
+	}
+	r := 0
```
stapelberg, line 120: Can you add a comment here about `m.Sglist[0]` containing the header (enforced by `Append()`), so that readers don’t need to wonder why `OutMessageHeaderSize` isn’t counted here?

internal/buffer/out_message.go
```diff
@@ -69,25 +53,15 @@ func (m *OutMessage) OutHeader() *fusekernel.OutHeader {
 // insufficient space, it returns nil.
 func (m *OutMessage) Grow(n int) unsafe.Pointer {
 	p := m.GrowNoZero(n)
```
stapelberg, line 55: Since there is no difference anymore between `Grow` and `GrowNoZero`, let’s unify them into one single method.

It would also be good to rename the method for clarity, since there is no growing going on anymore. Perhaps `AllocateAndAppend`?

fuseops/ops.go
stapelberg: Can you (perhaps in a separate PR) merge them back so that we can see what that looks like? In particular, changes to the intreface (fuseutil/file_system.go) also break compatibility for users.

If we end up going with two separate types, we need to factor out the common fields into a separate struct.

connection.go
```diff
@@ -505,10 +505,16 @@ func (c *Connection) Reply(ctx context.Context, opErr error) {
 	noResponse := c.kernelResponse(outMsg, inMsg.Header().Unique, op, opErr)
 
 	if !noResponse {
-		err := c.writeMessage(outMsg.Bytes())
+		var err error
+		if outMsg.Sglist != nil {
```
stapelberg, line 509: Could we alternatively move the initialization code from `Append()` somewhere else so that all messages have an Sglist that contains at least the header? Then, we don’t need to distinguish here and can remove the `Bytes()` method entirely

samples/readbenchfs/readbenchfs.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/jacobsa/fuse/fuseops"
+	"github.com/jacobsa/fuse/fuseutil"
+)
+
+type readBenchFS struct {
+	fuseutil.NotImplementedFileSystem
+	buf    []byte
+}
+
+const FILE_SIZE = 1024*1024*1024*1024
```
stapelberg, line 33: Constants shouldn’t be all-upper-case in Go. Make this `const fileSize`

stapelberg, line 33: Please add a comment like `// 1 TB`

fuseutil/not_implemented_file_system.go
```diff
@@ -138,6 +138,12 @@ func (fs *NotImplementedFileSystem) ReadFile(
 	return fuse.ENOSYS
 }
 
+func (fs *NotImplementedFileSystem) VectoredRead(
```
stapelberg, line 141: I know you’re just copying existing style, but here and elsewhere, can you please remove all line breaks from function signatures for newly added code? They should just be one line going forward.

stapelberg (commented):
Sorry for the long wait. Finally found a chance to try out this code and it seems to work, which is always a good start :) Left a few comments to make the code easier to read/maintain.

vitalif:
Hi, thanks for you comments, I'll try to review and fix them and update the patch shortly :-)

samples/readbenchfs/readbenchfs.go
vitalif: Done

connection.go
vitalif: I'm not sure if it really matters, but the idea here was to avoid an extra allocation of slice :-)

internal/buffer/out_message.go
vitalif: Ok, merged, I think Grow is fine, because it still grows the message

internal/buffer/out_message.go
vitalif: Ok, added comments

fuseutil/not_implemented_file_system.go
vitalif: I did it for readbenchfs, but not for new functions surrounded with the old code-style functions - I think a single function in a different style may look bad... Maybe we'll just change style in a separate commit and then change the style for newer commits?

vitalif:
OK, I fixed most of the issues, I'll try to make another version of the same patch without a separate VectoredReadOp to compare implementations and submit it

stapelberg:
Closing in favor of https://github.com/jacobsa/fuse/pull/120

PR #105, opened 2021-08-30
Add configuration to control async reads

mcdhee-msft:
In current jacobsa async reads is disabled for all users and nobody can switch it on if necessary. In this PR, I have created a configurable option in MountConfig to allow the user to control this feature

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -162,7 +162,7 @@ func (c *Connection) Init() error {
 	// Tell the kernel not to use pitifully small 4 KiB writes.
 	initOp.Flags |= fusekernel.InitBigWrites
 
-	if !c.cfg.DisableAsyncReads {
+	if c.cfg.EnableAsyncReads {
 		initOp.Flags |= fusekernel.InitAsyncRead
 	}
 
--- a/connection.go
+++ b/connection.go
@@ -161,6 +161,11 @@ func (c *Connection) Init() error {
 
 	// Tell the kernel not to use pitifully small 4 KiB writes.
 	initOp.Flags |= fusekernel.InitBigWrites
+
+	if !c.cfg.DisableAsyncReads {
+		initOp.Flags |= fusekernel.InitAsyncRead
+	}
+
 	// kernel 4.20 increases the max from 32 -> 256
 	initOp.Flags |= fusekernel.InitMaxPages
 	initOp.MaxPages = 256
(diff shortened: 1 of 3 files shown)


mount_config.go
```diff
@@ -174,6 +174,10 @@ type MountConfig struct {
 	// /proc/mounts will show the filesystem type as fuse.<Subtype>.
 	// If not set, /proc/mounts will show the filesystem type as fuse/fuseblk.
 	Subtype string
+
+	//Flag to disable async reads that are received from
```
stapelberg, line 178: Please add a space after each `//`

mount_config.go
```diff
@@ -174,6 +174,10 @@ type MountConfig struct {
...
+	//the kernel
+	EnableAsyncReads bool
```
stapelberg, line 180: The field is named EnableAsyncReads, but the comment says “Flag to disable”. Update the comment please.

connection.go
```diff
@@ -161,6 +161,11 @@ func (c *Connection) Init() error {
 
 	// Tell the kernel not to use pitifully small 4 KiB writes.
 	initOp.Flags |= fusekernel.InitBigWrites
+
+	if c.cfg.EnableAsyncReads {
```
stapelberg, line 165: When was this introduced in Linux?

What are the requirements on the client-side, if any, for users?

Can this be enabled by default for everyone or are there downsides?

stapelberg (commented):
Thanks for your PR!

connection.go
mcdhee-msft: https://github.com/jacobsa/fuse/commit/348ed9e731d013c43ea857ac6a9e1330e99faaa0

This feature was there in libfuse for quite sometime. But has been disabled for all only in jacobsa/fuse due to some errors that occurred in gcsfuse. But since jacobsa/fuse is an independent project. It is better if this option is configurable by the user.

mcdhee-msft: https://github.com/jacobsa/fuse/commit/348ed9e731d013c43ea857ac6a9e1330e99faaa0
From this link you can see the commit where this was disabled by default for everybody

mcdhee-msft: As Far as I know there isn't any downsides. But I think since this commit https://github.com/jacobsa/fuse/commit/d82237972ed9e7b2180ad27b3558f4c746a1fe16 this package seems to fail on rhel 7.5 onwards. But that is totally unrelated to this PR. Just a finding from our use case

stapelberg: > But I think since this commit [d822379](https://github.com/jacobsa/fuse/commit/d82237972ed9e7b2180ad27b3558f4c746a1fe16) this package seems to fail on rhel 7.5 onwards. But that is totally unrelated to this PR. Just a finding from our use case

Can you file an issue for this with details please?

stapelberg (approved):

PR #112, opened 2021-11-07
Run go test with GOMAXPROCS=1 to avoid hangs in CI

tetsuok:


stapelberg:
Why do things hang with the GOMAXPROCS default?

tetsuok:
This is still a draft pull request to see whether setting `GOMAXPROCS=1` helps identify cause of hangs or work around it in CI.

tetsuok:
It seems `go test` still hangs with `GOMAXPROCS=1`. I'll disable `go test` in the CI pipeline for Linux.

PR #115, opened 2021-11-07
Add gofmt CI step

tetsuok:
This adds a CI step to run `gofmt` to check that the code has been formatted by `gofmt`.


.github/workflows/ci.yml
```diff
@@ -9,6 +9,22 @@ on:
...
+    - name: Set up Go
+      uses: actions/setup-go@v2.1.4
+      with:
+        go-version: 1.16
+      id: go
+    # Check the codebase has been formatted by `gofmt` by checking whether
+    # the output of `gofmt -l` is empty, as recommended in
+    # https://github.com/golang/go/issues/24230)
+    - name: Run gofmt
+      run: test -z $(gofmt -l .)
```
stapelberg, line 26: Can you add another `gofmt -l` call here so that the output (unformatted files) makes it into the CI log?
Currently, this only shows *that* some files are not formatted, not *which* files are not formatted.

tetsuok: @stapelberg Sure. That's a good point. I actually added `gofmt -d` which prints diffs in addition to formatted files.

stapelberg (approved):

PR #116, opened 2021-11-07
Add a CI step to run go vet

tetsuok:
This adds a CI step to run `go vet` command which runs static analysis ([available checks](https://pkg.go.dev/cmd/vet)).


tetsuok:
This is still work in progress as we need to make sure all tests pass in Linux.

PR #119, opened 2022-01-04
refactor: Remove duplicate code in test

JakWai01:
I found a snipped of duplicate code in a test and removed it. The original copy is located right above. 

The unchanged code with the duplicate looks like this: 
```
// The offset should have been updated to point at the end of the file.
off, err = getFileOffset(f)
AssertEq(nil, err)
ExpectEq(13, off)

off, err = getFileOffset(f)
AssertEq(nil, err)
ExpectEq(13, off)
```


stapelberg (approved):

stapelberg:
Thanks!

PR #120, opened 2022-02-03
Vectored read - merged operation struct version

vitalif:
Hi, this is the other version of https://github.com/jacobsa/fuse/pull/104 PR
Maybe it really looks better, yeah :-)


stapelberg (approved):

stapelberg:
This seems to work and is indeed significantly faster, so let’s give it a shot! Thanks!

PR #123, opened 2022-05-26
Add support for the FUSE_BATCH_FORGET operation

ironsmile:
There are certain Kernel versions which do not send individual
Forget operations after they receive "not implemented" for Batch Forget.
One example is "5.4.0-110-generic" on Ubuntu 20.04. I am sure there are
plenty of others.

This leads to inode "leaks". Where reference counts are never decreased
and the inodes were left hanging around long after they are not needed
any more.

The best way to fix that was adding support for batch operations to the
lib. This way all users will be able to benefit from the batching
optimization.

### DIFF
--- a/fuseops/ops.go
+++ b/fuseops/ops.go
@@ -220,6 +220,32 @@ type ForgetInodeOp struct {
 	OpContext OpContext
 }
 
+// BatchForgetEntry represents one Inode entry to forget in the BatchForgetOp.
+//
+// Everything written in the ForgetInodeOp docs applies for the BatchForgetEntry
+// too.
+type BatchForgetEntry struct {
+	// The inode whose reference count should be decremented.
+	Inode InodeID
+
+	// The amount to decrement the reference count.
+	N uint64
+}
+
+// Decrement the reference counts for a list of inode IDs previously issued by the file
+// system.
+//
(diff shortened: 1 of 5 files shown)


fuseops/ops.go
```diff
@@ -220,6 +220,32 @@ type ForgetInodeOp struct {
 	OpContext OpContext
 }
 
+// BatchForgetEntry represents one Inode entry to forget in the BatchForgetOp.
+//
+// Everything written in the ForgetInodeOp docs applies for the BatchForgetEnetry
```
stapelberg, line 225: typo: BatchForgetEntry

stapelberg (changes requested):
Thanks for your PR! Can you rebase it onto current master? I pushed a commit to fix CI on mac in the meantime.

fuseops/ops.go
ironsmile: Well spotted! Fixed.

ironsmile:
You are welcome! I've rebased it to the latest master from 9 minutes ago too. Hope it builds now 😄

stapelberg:
Thanks!

vitalif:
Thanks, I did just the same thing yesterday and today I discovered that it's already done here :))

One thing I want to add is #126

PR #124, opened 2022-05-30
Support unprivileged mounts on Linux

blinsay:
Adds support for unprivileged mounts by allowing a parent process to pass an open `/dev/fuse` file descriptor as a mount point. Follows [the libfuse3 format](https://github.com/libfuse/libfuse/blob/a56147d3cb269d64635408ab5588ad66eba43943/ChangeLog.rst#libfuse-330-2018-11-06) and uses `/dev/fd/N` to represent an open fd.

I added a quick test for parsing `/dev/fd/N` fds. The `memfs` tests on master seem to hang and fail pretty consistently on my system.

### DIFF
--- a/mount_linux.go
+++ b/mount_linux.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
+	"strings"
 	"syscall"
 
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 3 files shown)


mount_linux.go
```diff
@@ -123,3 +133,19 @@ func mount(dir string, cfg *MountConfig, ready chan<- error) (*os.File, error) {
 	}
 	return dev, err
 }
+
+var DEV_FD = regexp.MustCompile("^/dev/fd/([0-9]+)$")
+
+func parseFuseFd(dir string) (int, error) {
+	matches := DEV_FD.FindAllStringSubmatch(dir, 1)
```
stapelberg, line 140: Can you use `strings.HasPrefix` and `strings.TrimPrefix` instead? All the regexp does is ensure the prefix matches — the `[0-9]+` part of it is covered by `strconv.Atoi` already.

stapelberg (changes requested):
Thanks for the PR! It looks good overall.

mount_linux.go
blinsay: The regexp didn't allow negative fds, so I ended up using `strconv.ParseUint` instead. Let me know how it looks.

blinsay: And thanks for the quick reply!

stapelberg (approved):

stapelberg:
Thanks!

PR #126, opened 2022-07-02
Fallback from BatchForgetOp to a series of ForgetInodeOp if unsupported

vitalif:
Hi, yesterday I implemented BatchForget support and then discovered it's already done in #123. However, one thing that I have to add is the fallback from a batch forget operation to multiple single-inode forget operations. This allows to enable batch forget operations transparently for all library users.


stapelberg (approved):

stapelberg:
Thanks!

vitalif:
Cool, thanks for quick reply :)

PR #127, opened 2022-07-15
Pass OpenFlags for OpenFileOp

xinmeigui-db:
Pass OpenFlags for OpenFileOp, so that we could have custom logic based on the operation.

### DIFF
--- a/fuseops/ops.go
+++ b/fuseops/ops.go
@@ -17,6 +17,8 @@ package fuseops
 import (
 	"os"
 	"time"
+
+	"github.com/jacobsa/fuse/internal/fusekernel"
 )
 
 ////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


fuseops/ops.go
```diff
@@ -646,6 +648,8 @@ type OpenFileOp struct {
 	// advance, for example, because contents are generated on the fly.
 	UseDirectIO bool
 
+	OpenFlags *fusekernel.OpenFlags
```
xinmeigui-db, line 651: I made this a pointer, so we don't default to `OpenReadOnly` (which is 0). Let me know if you have feedback.

stapelberg (approved):

stapelberg:
Thanks!

xinmeigui-db:
@stapelberg Thanks for the review! I was actually planning to make this a non-pointer, given it should always be set. Also was hoping to add more testing. If those sound good to you, I will open another PR with those changes.

stapelberg:
sure

xinmeigui-db:
@stapelberg I created https://github.com/jacobsa/fuse/pull/129. Please take a look when you get a chance, thanks!

PR #128, opened 2022-07-18
Avoid printing the error message related to 'no such file or directory'

avidullu:
Avoid printing the error message related to 'no such file or directory' on a writeMessage. These are mostly due to interacting with a file which does not exist, some customers report a large number of these errors in their logs. Internal issue id 236809802

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -39,17 +39,17 @@ var contextKey interface{} = contextKeyType(0)
 //
 // As of 2015-03-26, the behavior in the kernel is:
 //
-//  *  (http://goo.gl/bQ1f1i, http://goo.gl/HwBrR6) Set the local variable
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks for the PR. Can you explain under which circumstances this error occurs? (Ideally, explain how to reproduce the error with one of the sample file systems in this repo.)

avidullu:
Seems the error was happening a very niche case and we've added another check in gcsfuse. Closing this PR.

kahing:
@avidullu is there a reference for the gcsfuse issue? we are starting to see this as well when we updated our version of jacobsa/fuse

avidullu:
[Here](https://github.com/GoogleCloudPlatform/gcsfuse/pull/751) is the PR

kahing:
@avidullu did you ever figure out why those messages show up? hiding all fuse error isn't a great option

PR #129, opened 2022-07-18
Pass OpenFlags for OpenFileOp followups

xinmeigui-db:
Following up on https://github.com/jacobsa/fuse/pull/127/files with 2 changes:
1. do not make `OpenFlags` a pointer
2. add unit tests


samples/memfs/memfs.go
```diff
@@ -677,6 +681,20 @@ func (fs *memFS) OpenFile(
 		panic("Found non-file.")
 	}
 
+	if inode.name == CheckFileOpenFlagsFileName {
```
stapelberg, line 684: Is it required to put this here into the sample file system, or could you instead add a unit test?

Ideally the sample file system wouldn’t contain any test code, only example code.

samples/memfs/inode.go
```diff
@@ -33,6 +33,9 @@ type inode struct {
 	// Mutable state
 	/////////////////////////
 
+	// Name of the inode
```
stapelberg, line 36: Is this a relative name, as opposed to a path that can contain slashes? Maybe add an example.

samples/memfs/memfs.go
```diff
@@ -157,9 +163,10 @@ func (fs *memFS) getInodeOrDie(id fuseops.InodeID) *inode {
 //
 // LOCKS_REQUIRED(fs.mu)
 func (fs *memFS) allocateInode(
-	attrs fuseops.InodeAttributes) (id fuseops.InodeID, inode *inode) {
+	attrs fuseops.InodeAttributes,
```
stapelberg, line 166: Here and elsewhere, please don’t add line breaks into function signatures. I know the code currently does that, but it’s not idiomatic Go style and I’d like to get rid of line breaks whenever possible.

stapelberg (changes requested):

samples/memfs/memfs.go
xinmeigui-db: Yeah I agree this isn't ideal, but I could not think of a way to unit test it without changing the logic of the sample file system itself. Otherwise the unit test cannot know the value for `OpenFlags`. Suggestions appreciated, or maybe we can remove this if you prefer manual testing of the value of `OpenFlags`.

xinmeigui-db:
@stapelberg appreciate another review for this!

samples/memfs/memfs.go
stapelberg: Nevermind, for some reason I thought the samples are separate from tests, but now I see that the only _test.go files are in the samples directory. That’s a cleanup for another time, then…

stapelberg (approved):

PR #130, opened 2022-09-05
Add more debugging logs to the mounting process so we get more visibility for customer issues

avidullu:
Bugs like https://github.com/GoogleCloudPlatform/gcsfuse/issues/670 and https://github.com/GoogleCloudPlatform/gcsfuse/issues/715 are hard to debug because the mounting process is stuck at a very high level log and we are not able to get any more information about the underneath mounting process.
This PR adds some debug logs to the mount process so we can get some visibility into where the mounting gets stuck and help mitigate the issues.

### DIFF
--- a/mount.go
+++ b/mount.go
@@ -17,6 +17,7 @@ package fuse
 import (
 	"context"
 	"fmt"
+	"log"
 	"net"
 	"os"
 	"os/exec"
@@ -54,7 +55,7 @@ func Mount(
 
 	// Begin the mounting process, which will continue in the background.
 	if config.DebugLogger != nil {
-		config.DebugLogger.Println("Beginning the mounting process")
+		config.DebugLogger.Println("Beginning the mounting kickoff process")
 	}
 	ready := make(chan error, 1)
 	dev, err := mount(dir, config, ready)
@@ -62,7 +63,7 @@ func Mount(
 		return nil, fmt.Errorf("mount: %v", err)
 	}
 	if config.DebugLogger != nil {
-		config.DebugLogger.Println("Completed the mounting process")
+		config.DebugLogger.Println("Completed the mounting kickoff process")
 	}
 
 	// Choose a parent context for ops.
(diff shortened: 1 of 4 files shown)


mount.go
```diff
@@ -53,18 +53,27 @@ func Mount(
...
+	if config.DebugLogger != nil {
+		config.DebugLogger.Println("Beginning the mounting process")
+	}
 	ready := make(chan error, 1)
 	dev, err := mount(dir, config, ready)
 	if err != nil {
 		return nil, fmt.Errorf("mount: %v", err)
 	}
+	if config.DebugLogger != nil {
+		config.DebugLogger.Println("Completed the mounting process")
```
vadlakondaswetha, line 65: Here we are saying completed the mounting process and below we are saying waiting for mounting process to complete. Isn't is confusing.

mount_linux.go
```diff
@@ -108,6 +119,9 @@ func mount(dir string, cfg *MountConfig, ready chan<- error) (*os.File, error) {
 	// On linux, mounting is never delayed.
 	ready <- nil
 
+	if cfg.DebugLogger != nil {
+		cfg.DebugLogger.Println("Parsing fuse file descriptor")
+	}
```
vadlakondaswetha, line 124: should we add logs for fusermount also

mount_linux.go
```diff
@@ -84,6 +91,10 @@ func directmount(dir string, cfg *MountConfig) (*os.File, error) {
 	}
 	delete(opts, "subtype")
 	data += "," + mapToOptionsString(opts)
+
+	if cfg.DebugLogger != nil {
+		cfg.DebugLogger.Println("Starting the unix mounting")
+	}
```
vadlakondaswetha, line 97: should we add a log after unix.mount to mention that it returned. Most of the times we are seeing mounting started message but not completed message. This will help know if the mounting command is stuck.

mount.go
avidullu: Done.

mount.go
```diff
@@ -126,20 +127,34 @@ func checkMountPoint(dir string) error {
 	return nil
 }
 
-func fusermount(binary string, argv []string, additionalEnv []string, wait bool) (*os.File, error) {
+func fusermount(
+	binary string,
+	argv []string,
+	additionalEnv []string,
+	wait bool,
+	debugLogger *log.Logger) (*os.File, error) {
```
stapelberg, line 135: Please leave all the function parameters on one line. Splitting the function declaration is not common in Go.

avidullu: Done

stapelberg (approved):

mount.go
stapelberg: Yes, I know there are many pre-existing issues in this codebase.

gofmt doesn’t flag/fix all style issues, there’s also https://staticcheck.io/ for example.

stapelberg:
I updated the code to make the gofmt CI check pass. Can you rebase your PR please?

avidullu:
Done. Thanks for the help!

PR #131, opened 2022-09-23
Remove memfs PID check

djschaap:
Remove PID zero check from memfs entirely.

<STRIKE>Add allowPidZero argument to NewMemFS.</STRIKE>

memfs will now accept FUSE calls/operations from the kernel where PID is zero (instead of aborting, as it did previously).  This can happen when using multiple PID namespaces (i.e., mount propagation across containers).

Reference: https://github.com/libfuse/libfuse/issues/67

### DIFF
--- a/samples/memfs/memfs.go
+++ b/samples/memfs/memfs.go
@@ -200,10 +200,6 @@ func (fs *memFS) StatFS(
 func (fs *memFS) LookUpInode(
 	ctx context.Context,
 	op *fuseops.LookUpInodeOp) error {
-	if op.OpContext.Pid == 0 {
-		return fuse.EINVAL
-	}
-
 	fs.mu.Lock()
 	defer fs.mu.Unlock()
 
@@ -234,10 +230,6 @@ func (fs *memFS) LookUpInode(
 func (fs *memFS) GetInodeAttributes(
 	ctx context.Context,
 	op *fuseops.GetInodeAttributesOp) error {
-	if op.OpContext.Pid == 0 {
-		return fuse.EINVAL
-	}
-
 	fs.mu.Lock()
 	defer fs.mu.Unlock()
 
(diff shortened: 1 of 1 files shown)


stapelberg (commented):
I don’t mind the change, but can you explain for which purpose you need this? Are you running the tests in a namespace…?

djschaap:
Yes, that's it exactly.  I'm developing a FUSE client that runs within a container to mount a remote filesystem.  That mounted filesystem is then made available to several other containers.  As part of this, before developing my custom backend, I'm using some of the sample backends to validate & familiarize myself with basic FUSE functionality.

This pid zero check is a show-stopper that seems to have no benefit to my use.  Other sample filesystems (`hellofs`, specifically) seem to be fine without that check.

If you'd prefer not to merge this, that's fine with me - I can operate with a personal fork just fine.  I'm offering this PR with the hope it will be useful to someone else, someday.

And - thank you for maintaining this project!

stapelberg:
> This pid zero check is a show-stopper that seems to have no benefit to my use. Other sample filesystems (`hellofs`, specifically) seem to be fine without that check.

These checks were introduced in https://github.com/jacobsa/fuse/commit/36e01f1b3aaafad29231bd7286dac6c384110502 (cc @xinmeigui-db), but seem to be more for illustration of newly introduced pid field than out of any necessity.

Instead of making the check configurable, how about we just remove it entirely?

xinmeigui-db:
> 

cc @kahing , we are okay with removing it entirely.

xinmeigui-db:
> > This pid zero check is a show-stopper that seems to have no benefit to my use. Other sample filesystems (`hellofs`, specifically) seem to be fine without that check.
> 
> These checks were introduced in [36e01f1](https://github.com/jacobsa/fuse/commit/36e01f1b3aaafad29231bd7286dac6c384110502) (cc @xinmeigui-db), but seem to be more for illustration of newly introduced pid field than out of any necessity.
> 
> Instead of making the check configurable, how about we just remove it entirely?

cc @kahing , we are okay with removing it entirely.

djschaap:
That's great!  I will update this PR to remove the memfs pid check entirely.

stapelberg (approved):

PR #132, opened 2022-10-10
Pull request for some Get/Set set-UID/set-GID/stick-bit attributes + UID/GID context of the userland processus invoking syscall

folays:
Hello, what do you think of this PR ?

It has been tested on Mac & Linux (on Linux where it worked in production for at least 1 year+)

Those are needed to be able to 100% pass a POSIX test suite at least on Linux (I used https://github.com/pjd/pjdfstest) 

This PR adds ability :
- to "receive" requests to change set-UID/set-GID/sticky-bit `(chmod u+s / g+s / +t)` in `.SetInodeAttributes()`
- to figure out the "context" (UID/GID) of the user land processus invoking a syscall
(I think I probably needed this one to have newly created `inode` be POSIX-compliant (have the correct `uid:gid` and/or maybe some stick-bit group folder stuff, I don't remember)

Mac : I didn't verified the correctness or meaningfulness of those fields, but at least it doesn't prevent anything from working.
Linux : It works as expected.
Windows : Didn't tested at all.

### DIFF
--- a/mounted_file_system.go
+++ b/mounted_file_system.go
@@ -14,7 +14,10 @@
 
 package fuse
 
-import "context"
+import (
+	"context"
+	"fmt"
+)
 
 // MountedFileSystem represents the status of a mount operation, with a method
 // that waits for unmounting.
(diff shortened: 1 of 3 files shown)


mounted_file_system.go
```diff
@@ -47,3 +50,17 @@ func (mfs *MountedFileSystem) Join(ctx context.Context) error {
 		return ctx.Err()
 	}
 }
+
+// GetFuseContext implements the equiv. of FUSE-C get_fuse_context() and thus
```
stapelberg, line 54: The C equivalent is fuse_get_context, not get_fuse_context

stapelberg (changes requested):
Looks good overall

mounted_file_system.go
folays: Indeed. I forced-push on the branch of my PR to correct this.

stapelberg:
Thanks!

PR #134, opened 2022-11-18
Add device number support, publish file mode conversion functions

vitalif:
Hi

1) I recently discovered that the library lacked device number support and added it.
2) File modes used by Golang and GeeseFS are different so it's handy for file systems to have library conversion functions so I published them.

### DIFF
--- a/conversions.go
+++ b/conversions.go
@@ -209,6 +209,7 @@ func convertInMessage(
 			Parent: fuseops.InodeID(inMsg.Header().Nodeid),
 			Name:   string(name),
 			Mode:   convertFileMode(in.Mode),
+			Rdev:   in.Rdev,
 			OpContext: fuseops.OpContext{
 				FuseID: inMsg.Header().Unique,
 				Pid:    inMsg.Header().Pid,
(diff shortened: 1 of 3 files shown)


conversions.go
```diff
@@ -987,6 +965,38 @@ func convertFileMode(unixMode uint32) os.FileMode {
 	return mode
 }
 
+func ConvertGolangMode(inMode os.FileMode) uint32 {
```
stapelberg, line 968: The language is called Go, not Golang. Please rename to ConvertGoMode.

conversions.go
```diff
@@ -954,7 +932,7 @@ func convertChildInodeEntry(
 	convertAttributes(in.Child, &in.Attributes, &out.Attr)
 }
 
-func convertFileMode(unixMode uint32) os.FileMode {
+func ConvertFileMode(unixMode uint32) os.FileMode {
```
stapelberg, line 935: Here and below, please document all exported functions. Perhaps something like:

// ConvertFileMode returns an os.FileMode with the Go mode and permission bits set according to the Linux mode and permission bits.

stapelberg (changes requested):
Thanks for your PR!

stapelberg (approved):

PR #136, opened 2023-01-24
Upgrading go from 1.18 to 1.19

sethiay:
These changes are required for upgrading go to 1.19 in GCSFuse: https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/go.mod


stapelberg (approved):

stapelberg:
I don’t mind this change, but I don’t think it’s necessary either? What error message are you getting without it?

PR #137, opened 2023-02-18
Expose FuseID via OpContext

djschaap:
Would you be open to including the unique FuseID in the OpContext passed to every FUSE operation?

This would allow downstream functions/methods to include that number in detailed log messages, tying those messages back to the `Op` identifier in existing `debugLogger` output.  As it is, when multiple calls are active, correlating the messages can be difficult.

### DIFF
--- a/conversions.go
+++ b/conversions.go
@@ -51,15 +51,21 @@ func convertInMessage(
 		}
 
 		o = &fuseops.LookUpInodeOp{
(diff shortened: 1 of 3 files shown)


stapelberg (approved):

stapelberg:
Seems reasonable. Thanks!

PR #139, opened 2023-03-09
Adding handlePanic method 

Tulsishah:
Before, when an application crashed in the background, it was not getting logs anywhere. I made handlePanic method which will capture the panic and print panic messages in logfile

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -59,7 +59,7 @@ const maxReadahead = 1 << 20
 type Connection struct {
 	cfg         MountConfig
 	debugLogger *log.Logger
-	errorLogger *log.Logger
+	ErrorLogger *log.Logger
 
 	// The device through which we're talking to the kernel, and the protocol
 	// version that we're using to talk to it.
@@ -99,7 +99,7 @@ func newConnection(
 	c := &Connection{
 		cfg:         cfg,
 		debugLogger: debugLogger,
-		errorLogger: errorLogger,
+		ErrorLogger: errorLogger,
 		dev:         dev,
 		cancelFuncs: make(map[uint64]func()),
 	}
@@ -441,7 +441,7 @@ func (c *Connection) shouldLogError(
 	}
 
 	// We can't log if there's nothing to log to.
-	if c.errorLogger == nil {
+	if c.ErrorLogger == nil {
 		return false
 	}
 
@@ -503,7 +503,7 @@ func (c *Connection) Reply(ctx context.Context, opErr error) {
 
 	// Error logging
 	if c.shouldLogError(op, opErr) {
-		c.errorLogger.Printf("%T error: %v", op, opErr)
+		c.ErrorLogger.Printf("%T error: %v", op, opErr)
 	}
 
 	// Send the reply to the kernel, if one is required.
@@ -516,8 +516,8 @@ func (c *Connection) Reply(ctx context.Context, opErr error) {
 		} else {
 			err = c.writeMessage(outMsg.OutHeaderBytes())
 		}
-		if err != nil && c.errorLogger != nil {
-			c.errorLogger.Printf("writeMessage: %v %v", err, outMsg.OutHeaderBytes())
+		if err != nil && c.ErrorLogger != nil {
+			c.ErrorLogger.Printf("writeMessage: %v %v", err, outMsg.OutHeaderBytes())
 		}
 		outMsg.Sglist = nil
 	}
--- a/fuseutil/file_system.go
+++ b/fuseutil/file_system.go
@@ -129,6 +129,16 @@ func (s *fileSystemServer) handleOp(
 	c *fuse.Connection,
 	ctx context.Context,
 	op interface{}) {
+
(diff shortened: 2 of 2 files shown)


vadlakondaswetha (approved):

stapelberg:
This is unusual.

Panics go to STDERR by default. Why is STDERR not captured in your environment?

Even if the default panic behavior is not sufficient, why is the panic not recovered and logged at a higher level (in the program that uses jacobsa/fuse)?

Tulsishah:
> This is unusual.
> 
> Panics go to STDERR by default. Why is STDERR not captured in your environment?
> 
> Even if the default panic behavior is not sufficient, why is the panic not recovered and logged at a higher level (in the program that uses jacobsa/fuse)?

So in gcsfuse, we are using fuse library, and whenever we run gcsfuse in foreground mode and panic occurs, it is printing stderr on the console, which is not handled by gcsfuse. But when we ran gcsfuse in background mode, the stderr got vanished. And this stderr is not available anywhere in the system also. In gcsfuse all the operations are calling by this handleops method. That's why I made defer handlepanic call to print the panic message. Do you have any lead on achieving the same goal, or do you have any alternate ideas?

stapelberg:
> But when we ran gcsfuse in background mode, the stderr got vanished. And this stderr is not available anywhere in the system also

Can you describe how I can run gcsfuse in background mode? It’s surprising to me that STDERR is lost in this situation.

How can I reproduce the panic that you’re trying to diagnose here?

Tulsishah:
> > But when we ran gcsfuse in background mode, the stderr got vanished. And this stderr is not available anywhere in the system also
> 
> Can you describe how I can run gcsfuse in background mode? It’s surprising to me that STDERR is lost in this situation.
> 
> How can I reproduce the panic that you’re trying to diagnose here?

I put panic("message") in [createObject method](https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/internal/storage/bucket_handle.go#L106) and then use gcsfuse to create object

stapelberg:
Okay. The reason why the STDERR vanishes is because the daemonize code doesn’t set cmd.Stderr. When adding it like so, the panic is printed:

```patch
--- i/vendor/github.com/jacobsa/daemonize/daemonize.go
+++ w/vendor/github.com/jacobsa/daemonize/daemonize.go
@@ -211,6 +211,7 @@ func startProcess(
        cmd.Args = append(cmd.Args, args...)
        cmd.Env = append(cmd.Env, env...)
        cmd.ExtraFiles = []*os.File{pipeW}
+       cmd.Stderr = os.Stderr
 
        // Change working directories so that we don't prevent unmounting of the
        // volume of our current working directory.
```

Looking at the stack trace of a panic, I see that you already have an error-handling related wrapper:

```
github.com/googlecloudplatform/gcsfuse/internal/fs/wrappers.(*errorMapping).CreateFile(0xc000920180, {0x1153da8?, 0xc0009236b0?}, 0xc00074d710?)
	/home/michael/go/src/github.com/googlecloudplatform/gcsfuse/internal/fs/wrappers/error_mapping.go:164 +0x33
```

If you still need to recover panics for some reason, maybe you could recover panics in there?

Tulsishah:
> Okay. The reason why the STDERR vanishes is because the daemonize code doesn’t set cmd.Stderr. When adding it like so, the panic is printed:
> 
> ```diff
> --- i/vendor/github.com/jacobsa/daemonize/daemonize.go
> +++ w/vendor/github.com/jacobsa/daemonize/daemonize.go
> @@ -211,6 +211,7 @@ func startProcess(
>         cmd.Args = append(cmd.Args, args...)
>         cmd.Env = append(cmd.Env, env...)
>         cmd.ExtraFiles = []*os.File{pipeW}
> +       cmd.Stderr = os.Stderr
>  
>         // Change working directories so that we don't prevent unmounting of the
>         // volume of our current working directory.
> ```
> 
> Looking at the stack trace of a panic, I see that you already have an error-handling related wrapper:
> 
> ```
> github.com/googlecloudplatform/gcsfuse/internal/fs/wrappers.(*errorMapping).CreateFile(0xc000920180, {0x1153da8?, 0xc0009236b0?}, 0xc00074d710?)
> 	/home/michael/go/src/github.com/googlecloudplatform/gcsfuse/internal/fs/wrappers/error_mapping.go:164 +0x33
> ```
> 
> If you still need to recover panics for some reason, maybe you could recover panics in there?

In Background mode, where does it printed?

stapelberg:
> In Background mode, where does it printed?

To the terminal from which you started it in background mode.

If you start background mode from a systemd service, I would expect the output to make it into the system journal by default.

vitalif:
Capturing panics is bad, in fact I also have in https://github.com/yandex-cloud/geesefs left from Goofys but I plan to remove it. It only hides bugs. So I won't add panic handlers if you asked me

stapelberg:
Yeah, I think @Tulsishah found a better solution for this. I’ll close the PR.

Issue #143, opened 2023-05-11
Proposal to avoid memory copies between jacobsa FUSE library and application code


sbauersfeld:
Hello,

I work with a team of engineers and we are using a FUSE filesystem that uses the jacobsa library. Recently, we have been looking into performance improvements for reading and writing large files, and we found that copying data to the `ReadFile Dst` buffer and from the `WriteFile Data` buffer is reducing our performance. We created a prototype to remove these copies and saw up to 100-200 MB/s throughput performance improvements (~500 MB/s -> ~700 MB/s) for reads and (~400 -> ~600 MB/s) for writes.

The "vectored read" improvement added in [this PR](https://github.com/jacobsa/fuse/pull/120) is an important first step for skipping the data copy in `ReadFile`, but unfortunately it does not seem to fully satisfy our use case, because we don't have any way of knowing when the data has been successfully written back to the kernel (and thus, we don't know when we can free up the data we are holding in memory).

For `WriteFile`, we have a similar problem, because we want to return a response to the kernel as soon as possible, but we may not yet be ready to free up the `InMessage` containing the data for the `WriteFile` request.

I am open to suggestions here, but our proposal to satisfy our use cases for both `ReadFile` and `WriteFile` is to add an interface that would essentially wrap the `Freelists` that are used to get and free up `InMessage` and `OutMessage`.

For example, we would add an interface that looks something like:
`
type MessageProvider interface {

  func GetInMessage(size) *InMessage
  
  func GetOutMessage(size) *OutMessage
  
  func PutInMessage(InMessage)
  
  func PutOutMessage(OutMessage)

}
`

The mount configurations could allow a custom implementation of this interface to be provided, so that our application code can add hooks to both get notified when the In/OutMessage is no longer needed because the response has been written to the kernel, and we can also control when the In/OutMessage is actually freed up for future reuse.

If no custom implementation is provided in the mount configuration, then a default implementation of this interface would be a lightweight wrapper around the Freelists that exist today.

I am very happy to create a pull request for this issue. Do you have any particular thoughts and/or concerns before I do so?

Thanks,
Scott

stapelberg:
Hey Scott! That all seems reasonable to me. I haven’t looked much at the freelists TBH (they predate my involvement with the package), but we can hash out details over the actual PR. No thoughts or concerns beforehand :)

sbauersfeld:
Hi @stapelberg, I created a PR to address this issue: https://github.com/jacobsa/fuse/pull/144

[...]

PR #144, opened 2023-05-17
Introduce a MessageProvider interface so application code can control when message buffers are reused

sbauersfeld:
Follow up to https://github.com/jacobsa/fuse/issues/143

We (myself and a team of engineers) have been looking into performance improvements for reading and writing large files, and we found that copying data to the ReadFile Dst buffer and from the WriteFile Data buffer is reducing our performance. We created a prototype to remove these copies and saw up to 100-200 MB/s throughput performance improvements (~500 MB/s -> ~700 MB/s) for reads and (~400 -> ~600 MB/s) for writes.

The "vectored read" improvement added in https://github.com/jacobsa/fuse/pull/120 is an important first step for skipping the data copy in ReadFile, but unfortunately it does not seem to fully satisfy our use case, because we don't have any way of knowing when the data has been successfully written back to the kernel (and thus, we don't know when we can free up the data we are holding in memory).

For WriteFile, we have a similar problem, because we want to return a response to the kernel as soon as possible, but we may not yet be ready to free up the InMessage containing the data for the WriteFile request.

This PR proposes adding a MessageProvider interface so that we can add hooks to both get notified when `In/Out Messages` are being created and recycled, and also to control the behavior when necessary.

If no custom implementation is provided in the mount configuration, then a default implementation is provided that matches the existing behavior.


freelists.go
```diff
@@ -1,70 +0,0 @@
...
-
-	"github.com/jacobsa/fuse/internal/buffer"
-)
-
-////////////////////////////////////////////////////////////////////////
-// buffer.InMessage
-////////////////////////////////////////////////////////////////////////
-
-// LOCKS_EXCLUDED(c.mu)
-func (c *Connection) getInMessage() *buffer.InMessage {
```
sbauersfeld, line 28: The functions in this file were essentially just copied to `DefaultMessageProvider`

in_message.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package buffer
+package fuse
```
sbauersfeld, line 15: The InMessage and OutMessage classes need to be publicly available to be used along with the MessageProvider interface, so I propose moving these to the fuse package. Please let me know if you have other preferences

samples/memfs/memfs.go
```diff
@@ -627,7 +627,7 @@ func (fs *memFS) OpenFile(
 		// Set attribute (name=fileOpenFlagsXattr, value=OpenFlags) to test whether
 		// we set OpenFlags correctly. The value is checked in test with getXattr.
 		value := make([]byte, 4)
-		binary.LittleEndian.PutUint32(value, uint32(op.OpenFlags))
+		binary.LittleEndian.PutUint32(value, uint32(op.OpenFlags)&syscall.O_ACCMODE)
```
sbauersfeld, line 630: The tests for the open flags were broken for me because the 0x8000 bit seemed to be implicitly set somewhere (this test failed against master branch before my other changes were applied). I think this corresponds to the O_EVTONLY flag, which is maybe related to how golang opens files for reading with EPOLL?

sbauersfeld:
Hi @stapelberg, I would really appreciate it if you could take a look at this PR when you get a chance. Thanks!

sbauersfeld:
Hi @stapelberg , I still would really appreciate it if you could take a look at this PR sometime!

in_message.go
stapelberg: I’d like to keep the fuse package as small as possible. 

Can you implement it the other way around? Move the MessageProvider interface into the buffer package?

samples/memfs/memfs.go
stapelberg: I can reproduce this here. I’m not sure either where the extra 0x8000 bit comes from :-/

It would be interesting to git bisect the Linux kernel and track down when this behavior changed.

stapelberg (changes requested):

stapelberg:
Finally found some time to fix my development environment and run your code, and it works for me. Once you address the package change (see review comment), I think we can merge this.

sbauersfeld:
Hi @stapelberg , I addressed your comment in this PR. However, I also put up a different PR that allows us to accomplish the same thing from our FUSE server but is a bit simpler. Could you please take a look and let me know if you prefer that approach as well? https://github.com/jacobsa/fuse/pull/147

stapelberg:
closing in favor of https://github.com/jacobsa/fuse/pull/147

PR #146, opened 2023-07-05
FUSE-T macOS support

macos-fuse-t:
FUSE-T is a user space implementation of FUSE for macOS which doesn't require kernel extensions (KEXT)
More details at https://github.com/macos-fuse-t/fuse-t

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -467,6 +467,8 @@ func (c *Connection) shouldLogError(
 	return true
 }
 
+var writeLock sync.Mutex
+
 // Reply replies to an op previously read using ReadOp, with the supplied error
 // (or nil if successful). The context must be the context returned by ReadOp.
 //
@@ -510,6 +512,10 @@ func (c *Connection) Reply(ctx context.Context, opErr error) {
 	noResponse := c.kernelResponse(outMsg, inMsg.Header().Unique, op, opErr)
 
 	if !noResponse {
+		// writev is not atomic
+		writeLock.Lock()
+		defer writeLock.Unlock()
+
 		var err error
 		if outMsg.Sglist != nil {
 			_, err = writev(int(c.dev.Fd()), outMsg.Sglist)
--- a/internal/buffer/in_message.go
+++ b/internal/buffer/in_message.go
@@ -53,24 +53,6 @@ func NewInMessage() *InMessage {
 	}
 }
 
-func readAll(r io.Reader, dest []byte) (int, error) {
-	offset := 0
-
-	for offset < len(dest) {
-		// read the remaining buffer
-		n, err := r.Read(dest[offset:])
-		if n == 0 && err == nil {
-			// remote fd closed
-			return n, err
-		}
-		offset += n
-		if err != nil {
-			return offset, err
-		}
-	}
-	return offset, nil
-}
-
 var readLock sync.Mutex
 
 func (m *InMessage) ReadSingle(r io.Reader) (int, error) {
@@ -78,13 +60,13 @@ func (m *InMessage) ReadSingle(r io.Reader) (int, error) {
 	defer readLock.Unlock()
 
 	// read request length
-	if _, err := readAll(r, m.storage[0:4]); err != nil {
+	if _, err := io.ReadFull(r, m.storage[0:4]); err != nil {
 		return 0, err
 	}
 
 	l := m.Header().Len
 	// read remaining request
-	if n, err := readAll(r, m.storage[4:l]); err != nil {
+	if n, err := io.ReadFull(r, m.storage[4:l]); err != nil {
 		return n, err
 	}
 	return int(l), nil
--- a/internal/buffer/in_message.go
+++ b/internal/buffer/in_message.go
@@ -17,6 +17,7 @@ package buffer
 import (
 	"fmt"
 	"io"
+	"sync"
 	"syscall"
 	"unsafe"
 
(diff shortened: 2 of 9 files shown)


connection.go
```diff
@@ -510,6 +512,10 @@ func (c *Connection) Reply(ctx context.Context, opErr error) {
 	noResponse := c.kernelResponse(outMsg, inMsg.Header().Unique, op, opErr)
 
 	if !noResponse {
+		// writev is not atomic
```
stapelberg, line 515: What do you mean? Why can there only be one writev syscall at a time? Does it only affect writev() or also writeMessage()

internal/buffer/in_message.go
```diff
@@ -52,11 +53,56 @@ func NewInMessage() *InMessage {
 	}
 }
 
+func readAll(r io.Reader, dest []byte) (int, error) {
```
stapelberg, line 56: How is this function different from io.ReadFull?

internal/buffer/in_message.go
```diff
@@ -52,11 +53,56 @@ func NewInMessage() *InMessage {
...
+
 // Initialize with the data read by a single call to r.Read. The first call to
 // Consume will consume the bytes directly after the fusekernel.InHeader
 // struct.
 func (m *InMessage) Init(r io.Reader) error {
-	n, err := r.Read(m.storage[:])
+
+	var n int
+	var err error
+	if fusekernel.IsPlatformFuseT {
```
stapelberg, line 100: Why the difference in behavior here? Can you explain what the difference in semantics are and why fuse-t differs?

samples/hellofs/hello_fs.go
```diff
@@ -218,7 +218,7 @@ func (fs *helloFS) ReadDir(
 
 	// Grab the range of interest.
 	if op.Offset > fuseops.DirOffset(len(entries)) {
-		return fuse.EIO
+		return nil
```
stapelberg, line 221: Can you explain why it’s safe to change the error return in these samples? Was it wrong to begin with? Doesn’t the change affect the existing implementations?

stapelberg (changes requested):
Thanks for your PR.

I can’t find the source of fuse-t anywhere? Are you offering to take care of user-reported issues with fuse-t in this repository?

connection.go
macos-fuse-t: I don't know about Linux implementation of writev but on macOS writev is definitely not atomic, meaning if there are two concurrent writev() calls the data might be written in the wrong order. I added a mutex because I routinely hit this issue during my testing. write() on the other hand is fully atomic.

internal/buffer/in_message.go
macos-fuse-t: Perhaps I missed ReadFull() function. I will redo the PR

samples/hellofs/hello_fs.go
macos-fuse-t: I'm pretty much sure the correct convention in FUSE is to return a nil response rather than an error. I guess that linux  FUSE implementation is a bit lenient in that case. I could do the same but it would require a special treatment of this specific use case which I'd rather not do. I think the better way is to return nil in the user code implementation

macos-fuse-t:
> Thanks for your PR.
> 
> I can’t find the source of fuse-t anywhere? Are you offering to take care of user-reported issues with fuse-t in this repository?

Yes, I will take care of any user-reported issues related to fuse-t. fuse-t behaves a little bit different than conventional fuse, some minor features won't work correctly by virtue of being implemented over NFS but the benefits over osxfuse are significant

stapelberg:
Sorry for letting this PR linger for so long — the last few months have been very busy. I’ll try to return to it soon.

stapelberg (approved):

sethiay:
> @sethiay As I stated previously writev() function is not atomic (at least not on macOS), meaning you get unpredictable results if there are several concurrent writev() calls. 

I think you are talking about exclusion of one writev call from the other call for consistency and not atomicity. Also, the locks being added doesn't affect the atomicity of writev call. I am also not sure about consistency guarantees and found this on web: https://unix.stackexchange.com/questions/341050/im-wondering-how-the-atomicity-of-writev-is-ensured. Also, before this PR, without the writeLocks, we (GCSFuse) haven't received any consistency issues.

> I suppose the regression is on Linux? 

Yes

> Are you sure that it happens consistently because the lock looks quite innocuous. Perhaps the lock should be limited to macOS only if the regression is real. 

Yes, I ran the performance tests multiple times before flagging this issue. 

> Would you mind to tell me how do you test so I can try to reproduce it myself?

I don't have a minimal reproducer for the tests. However, I can explain the tests - The test runs 40 threads concurrently on a n2-standard 32 CPU cores Google Compute Engine VM where each thread continuously reads different 50MB file from a FUSE filesystem implementing jacobsa/fuse's filesystem interface for 60s.

macos-fuse-t:
@sethiay I'm looking into it. I'll try to send you a workaround to test. Please be patient because I'm a bit busy right now

stapelberg:
I don’t want to be too pushy, but could you provide a rough timeline estimate? Are we talking days, weeks, months?

I’m asking because @sethiay and team are eager to pick up https://github.com/jacobsa/fuse/commit/84480b28de63de1527052448f9b8fdd3ff6cbd33, which was submitted after the macOS changes.

If fixing this regression takes a few weeks, it might be best to revert the macOS changes for now and re-submit them later.

Thanks

macos-fuse-t:
I'll try to send a PR until the end of the week.

vitalif:
Hi, it seems that was a bad idea to make a new implementation the default one :) it broke mounting for everyone who used macfuse before. Also a user reported that geesefs crashes with fuse-t. I'll revert to macfuse by now ).

stapelberg:
> Hi, it seems that was a bad idea to make a new implementation the default one :) it broke mounting for everyone who used macfuse before. Also a user reported that geesefs crashes with fuse-t. I'll revert geesefs to macfuse by now ).

Can you send a PR to switch the default please?

PR #151, opened 2023-10-03
Handling error during initOp response to kernel

raj-prince:
Currently, we are not handling the error while responding the kernel. Ideally, we should handle this for all the operation but that requires more change. Currently, narrowing the scope to the error we get while communicating the InitOp response to kernel.

Related issues: https://github.com/GoogleCloudPlatform/gcsfuse/issues/583

Tested manually by adding a custom error message while writing message. Although I am not able to reproduce the above issue.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -193,8 +193,7 @@ func (c *Connection) Init() error {
 		initOp.Flags |= fusekernel.InitNoOpendirSupport
 	}
 
-	c.Reply(ctx, nil)
-	return nil
+	return c.Reply(ctx, nil)
 }
 
 // Log information for an operation with the given ID. calldepth is the depth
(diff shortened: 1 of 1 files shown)


connection.go
```diff
@@ -516,11 +515,16 @@ func (c *Connection) Reply(ctx context.Context, opErr error) {
 		} else {
 			err = c.writeMessage(outMsg.OutHeaderBytes())
 		}
-		if err != nil && c.errorLogger != nil {
-			c.errorLogger.Printf("writeMessage: %v %v", err, outMsg.OutHeaderBytes())
+		if err != nil {
+			if c.errorLogger != nil {
+				c.errorLogger.Printf("writeMessage: %v %v", err, outMsg.OutHeaderBytes())
+			}
+			return fmt.Errorf("writeMessage: %v %v", err, outMsg.OutHeaderBytes())
```
vadlakondaswetha, line 522: can you construct the error once and then use in the c.errorLogger.Print

stapelberg (approved):

Issue #122, opened 2022-05-26
Operations frequently canceled when accessing a `gcsfuse` mount with a Go application


axolotlgeoff:
I suspect what is happening:

1. A Go application accesses the mount
1. The application raises `SIGURG` signals due to a Go feature introduced in 1.14 for [non-cooperative goroutine preemption](https://github.com/golang/proposal/blob/master/design/24543-non-cooperative-preemption.md)
1. FUSE handles the signal and [raises an INTERRUPT](https://www.kernel.org/doc/html/latest/filesystems/fuse.html#interrupting-filesystem-operations):

    > If a process issuing a FUSE filesystem request is interrupted, the following will happen:
    > * If the request is not yet sent to userspace AND the signal is fatal (SIGKILL or unhandled fatal signal), then the request is dequeued and returns immediately.
    > * If the request is not yet sent to userspace AND the signal is not fatal, then an interrupted flag is set for the request. When the request has been successfully transferred to userspace and this flag is set, an INTERRUPT request is queued.
    > * If the request is already sent to userspace, then an INTERRUPT request is queued.

1. The `InterruptOp` is being [handled](https://github.com/jacobsa/fuse/blob/84920d11ddb7480758788f6cbb7718d96eda95ae/connection.go#L307) and cancels the operation
1. This cancels the context which is [passed to the HTTP request](https://github.com/jacobsa/gcloud/blob/d5f9e155f950dfa0ff2f30cc02e6ba684c7e8371/httputil/request.go#L63) to GCS resulting in errors such as:

    ```
    2022/05/25 15:20:21.417866 LookUpInode: operation canceled, clobbered: StatObject: not retrying StatObject("filename.example"): Get "https://storage.googleapis.com:443/storage/v1/b/example/o/filename.example?projection=full": net/http: request canceled
    fuse: 2022/05/25 15:20:21.417893 *fuseops.LookUpInodeOp error: operation canceled
    ```

Others, such as Docker and Gitea, have solved this by filtering out the `SIGURG` as referenced in https://github.com/golang/go/issues/37942 . However as far as I can tell there's no option to ignore the `SIGURG` since it's FUSE which is handling them, and by the time it reaches this library as an interrupt the context is lost.

Our current workaround is to set `GODEBUG="asyncpreemptoff=1"` for the applications which use `gcsfuse` mounts.

[...]

PR #153, opened 2023-12-08
Add config option to ignore signal interrupts.

bjornleffler:
Being able to ignore SIGURG fixes a number of known issues in GCS Fuse, as well as issue #122 for jacobsa/fuse.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -304,6 +304,17 @@ func (c *Connection) finishOp(
 
 // LOCKS_EXCLUDED(c.mu)
 func (c *Connection) handleInterrupt(fuseID uint64) {
+	// Allow fuse library clients to ignore interupt signals.
+	// NOTE(bjornleffler): Golang issues lots of SIGURG signals, which are
+	// interpreted as interuption signals.
+	// https://github.com/golang/proposal/blob/master/design/24543-non-cooperative-preemption.md
+	// It would have been preferrable to ignore SIGURG, but I couldn't
+	// figure out how to do that. From kernel documentation (https://goo.gl/H55Dnr):
+	// "The userspace filesystem may ignore the INTERRUPT requests entirely"
+	if c.cfg.IgnoreInterrupts {
+		return
+	}
+
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
--- a/fuseops/filetype.go
+++ b/fuseops/filetype.go
@@ -1,0 +1,62 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fuseops
+
+import (
+	"fmt"
+)
+
+type Filetype int
+
+const (
+	NoFiletype        Filetype = 0
+	RegularFiletype            = 1
+	DirectoryFiletype          = 2
+	SymlinkFiletype            = 3
+)
+
+const (
+	// Type values used in GCS metadata.
+	NoFiletypeString        = "none"
+	RegularFiletypeString   = "file"
+	DirectoryFiletypeString = "directory"
+	SymlinkFiletypeString   = "symlink"
+)
+
+func ParseFiletype(value string) (Filetype, error) {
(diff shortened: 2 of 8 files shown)


stapelberg:
I’m not convinced that this is the right fix: if you just don’t cancel the contexts, pending operations will pile up over time. 

I’m also not sure I fully understand the actual problem yet, as cancelling the operation should be fine, and the next FUSE request after the interruption should just start another operation, no? Do you have a standalone reproducer for this issue? (I asked for one in issue #122 as well and never got one.)

bjornleffler:
I agree that this this isn't the perfect fix. That would be to stop all SIGURG signals being delivered in the first place, but I couldn't figure out how to do that:
- os.signal.Ignore(...) didn't mask the signal.
- If you ask to be notified with os.signal.Notify(), the signal is delivered twice: first in user space, then by the kernel fuse call.

Cancelling the context is not fine, as it causes user visible application failures. A simple and consistent way to reproduce this is to run "git clone <repo>" in a fuse managed directory.

I don't fully understand the issue either. This seems like a Golang thing, not interupts coming from the kernel. What's even more surprising is that the kernel delivers a large number of SIGURG signals. About 50-100 signals for a git checkout with 11 files. Only a handful of those signals result in errors, possibly because the operations had already completed?

bjornleffler:
Given how:
1. Golang binaries seems to send lots of SIGURG signals to themselves.
2. This is user facing breaking issue.
3. The linux kernel says it's fine to ignore signals:
https://goo.gl/H55Dnr
"The userspace filesystem may ignore the INTERRUPT requests entirely, ..."

I reckon that this is the easiest way to adress a user facing issue. I deliberately left this as an opt-in feature, so we can test it and see if it causes pending operations to pile up over time.

stapelberg:
> That would be to stop all SIGURG signals being delivered in the first place, but I couldn't figure out how to do that:

I think the best way is to run your program with the environment variable `GODEBUG="asyncpreemptoff=1"`

> Cancelling the context is not fine, as it causes user visible application failures. 

Okay, but why? That’s the bit I don’t understand. The kernel’s INTERRUPT messages are supposed to signal that the client is no longer asking for the request to be answered. Is that incorrect? If it’s correct, where does the failure come from?

> A simple and consistent way to reproduce this is to run "git clone " in a fuse managed directory.

And what are the steps to do that? Please, supply steps I can run on a clean checkout of jacobsa/fuse without anything else. In particular, I don’t want to deal with Google Cloud accounts, projects, setup, etc. — I think narrowing down this problem to a minimal reproducer will be very valuable in understanding this issue.

jacobsa:
For what it's worth, I agree with @stapelberg here: it seems like this change needs to come with a detailed discussion of the root cause and why this is the appropriate place to address it. I haven't seen anything in these discussions (or the Google-internal ones) that indicates anybody has tried deeply to understand the root cause.

With regard to the quote from the kernel (docs "The userspace filesystem may ignore the `INTERRUPT` requests entirely"), I believe this is saying "a file system is free to give a poor user experience by allowing operations to be uninterruptable and potentially continue forever even when the user doesn't want them to continue". Doing that to work around this problem seems like throwing out the baby with the bath water, especially when it seems like it should be possible to deal with this another way or at least understand why it must be dealt with this way.

Below is a similar reply I gave on an internal discussion, with some potential avenues for investigation.

> This is of course your call, not mine, but I feel like some more root causing is warranted here. It seems like gcsfuse is doing the correct thing according to the kernel documentation and according to posix: an interrupted file system request is returning EINTR to indicate that it has been interrupted. Changing that, especially with a magic tuning parameter, probably should not be done lightly. I.e. not without fully understanding the problem.
>
> So then the remaining mystery is why this comes up with these programs in particular and why with gcsfuse in particular. Personally I would be trying to answer questions like the following:
>
> * What is the actual root cause? What signal is being delivered, and why?
> * Why doesn't git display this problem with local file system drivers?
> * Do local file system drivers really not return EINTR on interrupt?
> * What do other fuse server implementations that *do* support interrupts do?
> * Why doesn't gcsfuse display this problem with other applications?
> * What does posix say git is supposed to do with EINTR?
> * What does git actually do with EINTR?
> * If git is not handling EINTR correctly (a common bug), can we just fix git?
> * Is there some out of band way to find the signal number?

bjornleffler:
I did a bit more digging today and I'm more confused than before. I still don't know why the operations are interrupted. The interrupt problem seems unrelated to the SIGURG signals.

> I think the best way is to run your program with the environment variable `GODEBUG="asyncpreemptoff=1"`

What we do know is that the AWS and Azure equivalents to GCSFuse don't intercept signals in the fuse code and therefore don't experience the user facing problems.

Not to self: The SIGURG signals stop if you set GODEBUG="asyncpreemptoff=1" from the shell, but it doesn't work if you try to use os.Setenv("GODEBUG", "asyncpreemptoff=1") from inside the binary.

Issue #154, opened 2023-12-21
Add Mount Config to  select use OsxFuse or Fuse-T


labulakalia:
https://github.com/jacobsa/fuse/blob/2b0495a1d5e09afe79719e772a98b175303b4920/mount_darwin.go#L409-L419

I need test both OsxFuse  and  Fuse-T

stapelberg:
Can you explain why you need to work with both?

Can you send a PR to introduce an option?

cc @macos-fuse-t

labulakalia:
@stapelberg i want to make some performance test for osxfuse and fuse-t，and free to choice

PR #155, opened 2023-12-24
add MountConfig FuseType

labulakalia:
#154

### DIFF
--- a/mount_config.go
+++ b/mount_config.go
@@ -171,7 +171,7 @@ type MountConfig struct {
 
 	// OS X only.
 	//
-	// FuseType choice FUSE impl By OsxFuse/Fuse-T,default is OsxFuse
+	// FuseType choice FUSE impl By Fuse-T/OsxFuse,default is Fuse-T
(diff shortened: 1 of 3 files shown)


mount_config.go
```diff
@@ -169,6 +169,11 @@ type MountConfig struct {
 	// default name involving the string 'osxfuse' is used.
 	VolumeName string
 
+	// OS X only.
+	//
+	// FuseType choice FUSE impl By OsxFuse/Fuse-T,default is OsxFuse
```
stapelberg, line 174: This changes behavior, because currently the default is FuseT. Can you change the default to match the current behavior?

stapelberg (changes requested):

mount_config.go
labulakalia: okay, I will change it

stapelberg:
Merged (with a few small adjustments) as commit https://github.com/jacobsa/fuse/commit/bc4b1b58bb605f42b02aae44c7692105a016b58c

PR #157, opened 2024-01-08
Fix a spelling error

shenxingwuying:


fuseops/ops.go
```diff
@@ -977,7 +977,7 @@ type FallocateOp struct {
 	// If Mode is 0x0, allocate disk space within the range specified
 	// If Mode has 0x1, allocate the space but don't increase the file size
 	// If Mode has 0x2, deallocate space within the range specified
-	// If Mode has 0x2, it sbould also have 0x1 (deallocate should not increase
+	// If Mode has 0x3, it should also have 0x1 (deallocate should not increase
```
stapelberg, line 980: Your suggested change is not correct. Of course “sboud” is a typo for “should”, but the 0x2 is not a typo.

The mode flags for fallocate are defined in https://github.com/torvalds/linux/blob/master/include/uapi/linux/falloc.h

The comment really means that when 0x2 (FALLOC_FL_PUNCH_HOLE) is set, 0x1 (FALLOC_FL_KEEP_SIZE) also has to be set. This is confirmed in line 443 of linux/Documentation/filesystems/porting.rst: https://github.com/torvalds/linux/blob/052d534373b7ed33712a63d5e17b2b6cdbce84fd/Documentation/filesystems/porting.rst?plain=1#L443

stapelberg (changes requested):

fuseops/ops.go
shenxingwuying: ok, thank you for your explanation.

PR #158, opened 2024-04-04
Add PARALLEL_DIROPS FUSE option to jacobsa/fuse

sethiay:
Add PARALLEL_DIROPS FUSE option to jacobsa/fuse. When enabled, this allows parallel lookups and readdir calls to filesystem server in user space from FUSE driver.

Consider the directory and file structure is flat i.e. 
mnt/ 
   - child1.txt
   - child2.txt
   - ...
   - childN.txt
  
When PARALLEL_DIROPS is not enabled:
if an application tries to stat the child files parallely, for the first time, the calls are serialized from FUSE driver to filesystem server in user space. When the same files are accessed second time, the calls are parallel.

When PARALLEL_DIROPS is enabled:
if an application tries to stat the child files parallely, the calls are parallel from FUSE driver to filesystem server in user space.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -193,6 +193,11 @@ func (c *Connection) Init() error {
 		initOp.Flags |= fusekernel.InitNoOpendirSupport
 	}
 
+	// Tell the Kernel to allow sending parallel lookup and readdir operations.
+	if c.cfg.EnableParallelDirOps {
+		initOp.Flags |= fusekernel.InitParallelDirOps
+	}
+
 	return c.Reply(ctx, nil)
 }
 
--- a/internal/fusekernel/fuse_kernel.go
+++ b/internal/fusekernel/fuse_kernel.go
@@ -266,6 +266,7 @@ const (
 	InitAsyncDIO         InitFlags = 1 << 15
 	InitWritebackCache   InitFlags = 1 << 16
 	InitNoOpenSupport    InitFlags = 1 << 17
+	InitParallelDirOps   InitFlags = 1 << 18
 	InitMaxPages         InitFlags = 1 << 22
 	InitCacheSymlinks    InitFlags = 1 << 23
 	InitNoOpendirSupport InitFlags = 1 << 24
--- a/mount_config.go
+++ b/mount_config.go
@@ -185,6 +185,11 @@ type MountConfig struct {
 	// Flag to enable async reads that are received from
 	// the kernel
 	EnableAsyncReads bool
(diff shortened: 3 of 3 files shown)


stapelberg:
> Add PARALLEL_DIROPS FUSE option to jacobsa/fuse. When enabled, this allows parallel lookups and readdir calls to filesystem server in user space from FUSE driver.

Have you verified that it is safe to enable parallel dirops? 

When following your reference link to the Linux kernel change, the change says parallel dirops are disabled by default to prevent breakage. I’m not sure if jacobsa/fuse is ready for parallel dirops or what needs changing.

sethiay:
> > Add PARALLEL_DIROPS FUSE option to jacobsa/fuse. When enabled, this allows parallel lookups and readdir calls to filesystem server in user space from FUSE driver.
> 
> Have you verified that it is safe to enable parallel dirops?
> 
> When following your reference link to the Linux kernel change, the change says parallel dirops are disabled by default to prevent breakage. I’m not sure if jacobsa/fuse is ready for parallel dirops or what needs changing.

1. There is no [logic](https://github.com/jacobsa/fuse/blob/master/fuseutil/file_system.go#L144) on jacobsa/fuse side specifically for LookUpInode or Readdir FUSE driver calls. 
2. The PR changes doesn't make allow parallel lookups by default. So, if a wrapper over jacobsa/fuse changes it they will be aware of it and handle parallel lookups accordingly.
3. Also in my experiment without the PR changes, (a) if files are nested in directories and if parallel lookups are done on different directories then FUSE driver sends parallel calls. (b) on repeated parallel lookups of files under same directories, FUSE driver sends parallel calls.

> the change says parallel dirops are disabled by default to prevent breakage. I’m not sure if jacobsa/fuse is ready for parallel dirops or what needs changing.
Given above that we are not making this true by default and the parallel look ups should be handled by wrapper over jacobsa/fuse implementing filesystem methods, so we should be good here ?
go-fuse also supports this option: https://github.com/hanwen/go-fuse/blob/2901efa04e040ebe54a951bcea93d67ab1af1ad5/fuse/types.go#L297
libfuse also support this option: https://github.com/libfuse/libfuse/blob/e2df577a7ca0c4b06d9eb36eb79138e34cc03088/ChangeLog.rst#libfuse-300-2016-12-08

stapelberg:
Thanks for the background!

stapelberg (approved):

PR #160, opened 2024-05-08
Add SyncFS support

socketpair:


internal/fusekernel/fuse_kernel.go
```diff
@@ -46,7 +46,7 @@ const (
 	ProtoVersionMinMajor = 7
 	ProtoVersionMinMinor = 18
 	ProtoVersionMaxMajor = 7
-	ProtoVersionMaxMinor = 31
+	ProtoVersionMaxMinor = 34
```
socketpair, line 49: I'm unsure it is required. I'm totally new in FUSE.

fuseutil/file_system.go
```diff
@@ -238,5 +239,9 @@ func (s *fileSystemServer) handleOp(
 		err = s.fs.Fallocate(ctx, typed)
 	}
 
+	case *fuseops.SyncFSOp:
+		err = s.fs.SyncFS(ctx, typed)
+	}
```
vitalif, line 244: extra }

socketpair: fixed

stapelberg (approved):

PR #161, opened 2024-05-09
Fix IO performance regression 

macos-fuse-t:
Following @sethiay regression report, restrict writev locking to macOS fuse-t platform. 
Also corrects an issue with special folders handling such as "Google Drive" in loopbackfs.


stapelberg:
Thanks!

PR #167, opened 2024-10-21
Allow filesystems to set atomic_o_trunc flag

sbauersfeld:
This PR implements the atomic_o_trunc flag during the fuse init protocol.

When this flag is enabled, application calls to `open` with the O_TRUNC flag will cause a FUSE OpenFile op with the O_TRUNC flag set. In comparison, the default behavior is an OpenFile op without O_TRUNC, followed by a SetInodeAttributes op with the target size set to 0.

Ref: https://github.com/torvalds/linux/commit/6ff958edbf39c014eb06b65ad25b736be08c4e63

### DIFF
--- a/samples/memfs/memfs.go
+++ b/samples/memfs/memfs.go
@@ -640,7 +640,7 @@ func (fs *memFS) OpenFile(
 		// Set attribute (name=fileOpenFlagsXattr, value=OpenFlags) to test whether
 		// we set OpenFlags correctly. The value is checked in test with getXattr.
 		value := make([]byte, 4)
-		binary.LittleEndian.PutUint32(value, uint32(op.OpenFlags)&syscall.O_ACCMODE)
+		binary.LittleEndian.PutUint32(value, uint32(op.OpenFlags))
 		err := fs.setXattrHelper(inode, &fuseops.SetXattrOp{
 			Name:  FileOpenFlagsXattrName,
 			Value: value,
--- a/connection.go
+++ b/connection.go
@@ -200,6 +200,10 @@ func (c *Connection) Init() error {
 		initOp.Flags |= fusekernel.InitParallelDirOps
 	}
 
+	if c.cfg.EnableAtomicTrunc {
+		initOp.Flags |= fusekernel.InitAtomicTrunc
+	}
+
 	return c.Reply(ctx, nil)
 }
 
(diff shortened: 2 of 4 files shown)


sbauersfeld:
Hi @stapelberg, I'd appreciate your review on this change when you get a chance. I tested the new version of this library with our fuse filesystem and the atomic_o_trunc flag worked as expected. I also added some unit tests. Thanks, Scott

samples/memfs/memfs.go
```diff
@@ -640,7 +640,7 @@ func (fs *memFS) OpenFile(
 		// Set attribute (name=fileOpenFlagsXattr, value=OpenFlags) to test whether
 		// we set OpenFlags correctly. The value is checked in test with getXattr.
 		value := make([]byte, 4)
-		binary.LittleEndian.PutUint32(value, uint32(op.OpenFlags)&syscall.O_ACCMODE)
+		binary.LittleEndian.PutUint32(value, uint32(op.OpenFlags))
```
stapelberg, line 643: Huh, we had to add this over in PR #147 and now you say it is no longer needed? Can it be kept, or are you saying it must be removed? Can we figure out where this came from and what happened?

stapelberg (changes requested):

samples/memfs/memfs.go
sbauersfeld: Yeah, we had to add this mask in the other PR because the OpenFlags received by the memfs filesystem have some extra bit set (the actual value is 0x8000, 0x8001, or 0x8002 instead of just 0x0000, 0x0001, 0x0002). So even if the unit test is calling open with only the O_RDONLY flag set, memfs gets open flags 0x8000 instead of 0x0000.

Previously, the unit tests checked for exact matching between the flags passed to open by the unit test and the flags received by memfs. The test was failing because the extra 0x8000 bit seems to be passed to fuse by the kernel. I fixed this by adding the syscall.O_ACCMODE mask inside memfs OpenFile, but I know think this is a mistake since it does not allow us to test other flags passed to OpenFile (like O_TRUNC).

In this PR, I removed the mask from memfs and moved it to where the unit tests actually compare the open flags: https://github.com/jacobsa/fuse/pull/167/files#diff-ed9bb5c67338f67942527e1ad386d7331e1672678c951caf7cea60e9b33566c9R96

stapelberg (approved):

PR #170, opened 2025-02-24
Improving fuse error log for better debuggability

Tulsishah:
Currently, error logs lack operation IDs, which makes it difficult to correlate errors with specific requests during debugging. Adding the operation ID to the logs will significantly improve debugging efficiency. 

Furthermore, because debug responses and error logs are essentially the same, we have removed the `debug severity` response logs, retaining only the error logs. This will also reduce the overall number of logs.

Previously
```
{"timestamp":{"seconds":1740394276,"nanos":873104409},"severity":"TRACE","message":"fuse_debug: Op 0x00000166        connection.go:420] <- SyncFile (inode 2, PID 173511)"}
{"timestamp":{"seconds":1740394276,"nanos":897312524},"severity":"TRACE","message":"fuse_debug: Op 0x00000166        connection.go:513] -> Error: \"stale file handle\""}
{"timestamp":{"seconds":1740394276,"nanos":897321854},"severity":"ERROR","message":"fuse: *fuseops.SyncFileOp error: stale file handle"}
```

Now
```
{"timestamp":{"seconds":1740393686,"nanos":777589764},"severity":"TRACE","message":"fuse_debug: Op 0x00000168        connection.go:420] <- SyncFile (inode 2, PID 167380)"}
{"timestamp":{"seconds":1740393686,"nanos":808585082},"severity":"ERROR","message":"fuse: Op 0x00000168      *fuseops.SyncFileOp] -> Error: \"stale file handle\""}
```

More info: [b/398756683](https://b.corp.google.com/issues/398756683)

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -524,7 +524,7 @@ func (c *Connection) Reply(ctx context.Context, opErr error) error {
 
 	// Error logging
 	if c.shouldLogError(op, opErr) {
-		c.errorLogger.Printf("%T error: %v", op, opErr)
+		c.errorLogger.Printf("Op 0x%08x %T] -> Error: %q", fuseID, op, opErr)
 	}
 
 	// Send the reply to the kernel, if one is required.


connection.go
```diff
@@ -524,7 +524,13 @@ func (c *Connection) Reply(ctx context.Context, opErr error) error {
```
ashmeenkaur, line 524: should line 521 be an error log instead? why log the same error twice?

Tulsishah: If we make it error then simple "not exist" from LookUpInode error will also come as error severity. Which seems incorrect to me.

Tulsishah: Earlier thought to not make bigger change that's why added only op ID to existing log. but we can do it now also. Done.

ashmeenkaur (approved):

connection.go
```diff
@@ -518,15 +518,20 @@ func (c *Connection) Reply(ctx context.Context, opErr error) error {
 		if opErr == nil {
 			c.debugLog(fuseID, 1, "-> %s", describeResponse(op))
 		} else {
-			c.debugLog(fuseID, 1, "-> Error: %q", opErr.Error())
+			if c.shouldLogError(op, opErr) {
+				// Error logging
+				msg := fmt.Sprintf(
+					"Op 0x%08x %24s] %v",
```
abhishek10004, line 524: Can't we directly print the type by using %T here?

Tulsishah: It is not working as expected. Printing all the params instead of name. 
```
{"timestamp":{"seconds":1740467027,"nanos":396842858},"severity":"ERROR","message":"fuse: Op 0x0000018a &{%!s(fuseops.InodeID=                       1)              autorun.inf {%!s(fuseops.InodeID=                       0) %!s(fuseops.GenerationNumber=                       0) {%!s(uint64=                       0) %!s(uint32=                       0)               ---------- %!s(uint32=                       0) 0001-01-01 00:00:00 +0000 UTC 0001-01-01 00:00:00 +0000 UTC 0001-01-01 00:00:00 +0000 UTC 0001-01-01 00:00:00 +0000 UTC %!s(uint32=                       0) %!s(uint32=                       0)} 0001-01-01 00:00:00 +0000 UTC 0001-01-01 00:00:00 +0000 UTC} {%!s(uint64=                     394) %!s(uint32=                  701950) %!s(uint32=                 1052824)}}] string"}
```

abhishek10004: Thanks for checking.

abhishek10004 (approved):
LGTM

connection.go
stapelberg: I suspect you made a mistake while testing? I don’t see a reason why nested fmt.Sprintf calls should be needed here.

When I try this:

```go
			if c.shouldLogError(op, opErr) {
				// Error logging
				msg := fmt.Sprintf(
					"Op 0x%08x %T] -> Error: %q",
					fuseID, op, opErr)
				c.errorLogger.Println(msg)
			} else {
				c.debugLog(fuseID, 1, "-> Error: %q", opErr.Error())
			}
```

I see the following output:

```
ERR: Op 0x0000000e *fuseops.ReleaseDirHandleOp] -> Error: "function not implemented"
```

Please change the code to use a single format string

stapelberg (changes requested):
Looks good overall

connection.go
Tulsishah: Ohh yeah my bad. Fixed.

Tulsishah: Ohh yeah my bad. Fixed.

connection.go
```diff
@@ -518,15 +518,18 @@ func (c *Connection) Reply(ctx context.Context, opErr error) error {
 		if opErr == nil {
 			c.debugLog(fuseID, 1, "-> %s", describeResponse(op))
 		} else {
-			c.debugLog(fuseID, 1, "-> Error: %q", opErr.Error())
+			if c.shouldLogError(op, opErr) {
```
stapelberg, line 521: Now that you moved the error logging code into this branch of the conditional, that means that users will not see error logs when they don’t enable debug logging, which does not seem right to me.

It might be easier to split the error/debug log reduction into a separate PR to unblock adding the operation ID.

Tulsishah: Raised separate PR - https://github.com/jacobsa/fuse/pull/172

PR #176, opened 2025-04-30
Add support for detecting file open mode : append

anushka567:
This PR introduces a method to expose the file open mode specifically when the file is opened in append mode.

### DIFF
--- a/internal/fusekernel/fuse_kernel.go
+++ b/internal/fusekernel/fuse_kernel.go
@@ -199,6 +199,12 @@ func (fl OpenFlags) IsReadWrite() bool {
 	return fl&OpenAccessModeMask == OpenReadWrite
 }
 
+// Return true if OpenAppend is set.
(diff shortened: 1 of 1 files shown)


internal/fusekernel/fuse_kernel.go
```diff
@@ -199,6 +199,12 @@ func (fl OpenFlags) IsReadWrite() bool {
 	return fl&OpenAccessModeMask == OpenReadWrite
 }
 
+// Return true if OpenAppend is set.
+func (fl OpenFlags) IsAppend() bool {
+	flags:=fl &^ OpenAccessModeMask
```
stapelberg, line 204: I don’t think removing the access mode bits is necessary. Please remove this line, the function can be implemented as `return fl&OpenAppend == OpenAppend`

Also, I see that your code is not formatted. Please always use `gofmt` to format code before sending a PR. Thanks!

stapelberg (approved):

PR #177, opened 2025-05-14
Add support for inode and dentry invalidation.

cjc25:
Package fusekernel already had definitions of the NotifyInvalInodeOut and NotifyInvalEntryOut structures. We add a Notifier, which allows filesystem implementations to use those to invalidate kernel cache entries.

By using these invalidation mechanisms, user code can still mostly use the page cache, but with some feedback from the filesystem daemon when contents have changed dynamically.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -378,6 +378,23 @@ func (c *Connection) readMessage() (*buffer.InMessage, error) {
 	}
 }
 
+// Write a buffer.OutMessage to the kernel, with writev if vectored IO is useful
+// and write if not.
+func (c *Connection) writeOutMessage(outMsg *buffer.OutMessage) error {
+	var err error
+	if outMsg.Sglist != nil {
+		if fusekernel.IsPlatformFuseT {
+			// writev is not atomic on macos, restrict to fuse-t platform
+			writeLock.Lock()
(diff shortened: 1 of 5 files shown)


cjc25:
This is somewhat similar to #140 . Some comparisons:

- Instead of injecting the Connection into the FileSystem interface, this PR uses a wrapper Server, since ServeOps already had access to the Connection
- Instead of allowing any calls to Connection functions, this PR wraps the writes by requiring the use of an interface in package fuse to issue the writes/notifications
- #140 writes to the fd directly and risks nil dereference on Connection. This PR wraps the writes in channels to serialize them (probably unnecessary, but convenient to think about), and risks deadlock if an Invalidate function must return before ServeOps can be called (which would be Weird, IMO).
- Both PRs allow the fuse Server to deadlock if the user performs an invalidation while a "related operation" is outstanding.
- This PR has a test implementation :). It mirrors the similar test implementations in libfuse.

stapelberg (approved):

stapelberg:
Nice! Thanks for the clean PR. It passes CI, passes tests and seems to work in a quick manual test.

I am inclined to merge it.

@vitalif This should work for your use-case over in PR #140, too, right?

PR #178, opened 2025-06-04
Added Readdirplus support

aditimittal2003:
Integrate Readdirplus support in Jacoba Fuse

This commit introduces Readdirplus functionality into the Jacobsa Fuse.

Key changes in this PR include:

1. New ReadDirPlusOp Struct (fuse/fuseops/ops.go)

2. FileSystem Interface Update (fuseutil/file_system.go):
    Added the ReadDirPlus(context.Context, *fuseops.ReadDirPlusOp) error method to the FileSystem interface.

3. Operation Handling:
    * The handleOp function has been updated to include a case for fuseops.ReadDirPlusOp to dispatch to the new ReadDirPlus method.
    * The convertInMessage function now handles fusekernel.OpReaddirplus to convert the kernel message to the appropriate op. 
    * The kernelResponseForOp function now includes a case for fuseops.ReadDirPlusOp. 

4. NotImplementedFileSystem Update:
    * The ReadDirPlus method has been added, returning fuse.ENOSYS by default.  This serves as a placeholder until full implementation is enabled or for file systems not supporting it. 

5. Mount Configuration and Kernel Flags:
    * Added DisableReaddirplus to MountConfig (mount_config.go) to provide an option to disable the Readdirplus capability and revert to using ReadDir. 
    * Enabled fusekernel.InitDoReaddirplus in init flags, signaling to the kernel that GCSFuse can handle Readdirplus requests. 
    * Enabled fusekernel.InitReaddirplusAuto in init flags, allowing the kernel to adaptively choose between Readdirplus and Readdir based on its heuristics. 

These changes lay the groundwork in Jacobsa Fuse for GCSFuse to leverage Readdirplus for improved directory listing performance.

### DIFF
--- a/fusetesting/readdirplus.go
+++ b/fusetesting/readdirplus.go
@@ -15,153 +15,32 @@
 package fusetesting
 
 import (
-	"bufio"
-	"bytes"
 	"fmt"
-	"io/fs"
 	"os"
-	"os/exec"
-	"regexp"
 	"sort"
-	"strconv"
-	"strings"
-	"time"
 )
 
-// FileInfo is a custom implementation of the os.FileInfo interface.
-type FileInfo struct {
-	name    string
-	size    int64
-	mode    os.FileMode
-	modTime time.Time
-}
-
-func (fi FileInfo) Name() string       { return fi.name }
-func (fi FileInfo) Size() int64        { return fi.size }
-func (fi FileInfo) Mode() os.FileMode  { return fi.mode }
-func (fi FileInfo) ModTime() time.Time { return fi.modTime }
-func (fi FileInfo) IsDir() bool        { return fi.mode.IsDir() }
-func (fi FileInfo) Sys() interface{}   { return nil }
-
-// Regular Expression for parsing a long listing line
-var longListingRegex = regexp.MustCompile(
-	`^([dl-][rwxstST-]{9})` + // File Mode (e.g., -rw-r--r--, drwxr-xr-x)
-		`\s+(\S+)` + // Number of links
-		`\s+(\S+)` + // User
-		`\s+(\S+)` + // Group
-		`\s+(\d+)` + // Size
-		`\s+([A-Za-z]{3})` + // Month
-		`\s+(\d{1,2})` + // Day
-		`\s+(\d{2}:\d{2}|\d{4})` + // Time (HH:MM) or Year (YYYY)
-		`\s+(.+)$`, // File Name
-)
-
-// parseFileMode converts a permission string (e.g., "-rw-r--r--") to fs.FileMode.
-func parseFileMode(permissionStr string) (fs.FileMode, error) {
-	if len(permissionStr) != 10 {
-		return 0, fmt.Errorf("invalid permission string length: %q", permissionStr)
-	}
-
-	var fileMode fs.FileMode
-
-	// Set directory or symbolic link bit based on first character.
-	switch permissionStr[0] {
-	case 'd':
-		fileMode |= fs.ModeDir
-	case 'l':
-		fileMode |= fs.ModeSymlink
-	case '-':
-		// Regular file; no additional mode bits needed for type.
-	default:
-		return 0, fmt.Errorf("unknown file type character: %c in %q", permissionStr[0], permissionStr)
-	}
-
-	// Parse the 9 permission bits.
-	for i, r := range permissionStr[1:] {
-		if r != '-' {
-			// Calculate the permission bit (e.g., 'r' at index 1 maps to 0o400).
-			fileMode |= (1 << (8 - i))
-		}
-	}
-	return fileMode, nil
-}
-
-// ParseLine parses a single line of long listing output and returns a FileInfo struct.
-func parseLine(line string) (fs.FileInfo, error) {
-	match := longListingRegex.FindStringSubmatch(line)
-	if len(match) == 0 {
-		return nil, fmt.Errorf("invalid line format, no match found for: %q", line)
-	}
-
-	permissionStr := match[1]
-	sizeStr := match[5]
-	fileName := match[9]
-
-	// Parse File Mode
-	fileMode, err := parseFileMode(permissionStr)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse file mode from %q: %w", permissionStr, err)
-	}
-
-	// Parse File Size
-	fileSize, err := strconv.ParseInt(sizeStr, 10, 64)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse file size %q: %w", sizeStr, err)
-	}
-
-	return FileInfo{
-		name: fileName,
-		size: fileSize,
-		mode: fileMode,
-	}, nil
-}
-
-// ReadDirPlusPicky executes long listing command on the given directory and parses its output
-// into a sorted list of os.FileInfo objects.
+// Read the directory with the given name and returns a list of os.FileInfo entries,
+// sorted by name.
 func ReadDirPlusPicky(dirname string) (entries []os.FileInfo, err error) {
-	cmd := exec.Command("ls", "-l", dirname)
-	var stderr bytes.Buffer
-	cmd.Stderr = &stderr
-
-	stdout, err := cmd.StdoutPipe()
+	// Open the directory.
+	f, err := os.Open(dirname)
 	if err != nil {
-		return nil, fmt.Errorf("getting stdout pipe for ls command: %w", err)
+		return nil, fmt.Errorf("Open: %v", err)
 	}
 
-	if err := cmd.Start(); err != nil {
-		return nil, fmt.Errorf("starting ls command: %w", err)
-	}
-
-	// Read the output line by line.
-	scanner := bufio.NewScanner(stdout)
-	var lines []string
-	for scanner.Scan() {
-		lines = append(lines, scanner.Text())
-	}
-
-	if err := cmd.Wait(); err != nil {
-		if stderr.Len() > 0 {
-			errMsg := strings.ToLower(strings.TrimSpace(stderr.String()))
-			return nil, fmt.Errorf("ls failed with message %q: %w", errMsg, err)
+	// Don't forget to close it later.
+	defer func() {
+		closeErr := f.Close()
+		if closeErr != nil && err == nil {
+			err = fmt.Errorf("Close: %v", closeErr)
 		}
-		return nil, fmt.Errorf("waiting for ls command to finish: %w", err)
-	}
+	}()
 
-	// Skip the "total" line often included in the output.
-	if len(lines) > 0 && strings.HasPrefix(lines[0], "total") {
-		lines = lines[1:]
-	}
-
-	// Iterate through each line of the output and parse it.
-	for _, line := range lines {
-		if line == "" {
-			continue // Skip empty lines.
-		}
-		entry, err := parseLine(line)
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse line %q: %w", line, err)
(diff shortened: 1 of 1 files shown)


mount_config.go
```diff
@@ -205,6 +205,11 @@ type MountConfig struct {
 	// without O_TRUNC, followed by a SetInodeAttributes op with the target size set to 0.
 	// Ref: https://github.com/torvalds/linux/commit/6ff958edbf39c014eb06b65ad25b736be08c4e63
 	EnableAtomicTrunc bool
+
+	// Disable ReadDirPlus capability so ReadDir is used instead. Simple
+	// directory queries (i.e. 'ls' without '-l') can be faster with
+	// ReadDir
+	DisableReaddirplus bool
```
raj-prince, line 212: I feel, EnableReaddirplus seems better option, as plan is to keep it by default off. Accordingly, we need to change the do, explaining the benefit of this feature.

connection.go
```diff
@@ -204,6 +204,14 @@ func (c *Connection) Init() error {
 		initOp.Flags |= fusekernel.InitAtomicTrunc
 	}
 
+	if !c.cfg.DisableReaddirplus {
+		// Enable Readdirplus support, allowing the kernel to use Readdirplus
+		initOp.Flags |= fusekernel.InitDoReaddirplus
+
+		// Enable adaptive Readdirplus, allowing the kernel to choose between Readdirplus and Readdir
+		initOp.Flags |= fusekernel.InitReaddirplusAuto
```
raj-prince, line 212: I feel, we should expose ReaddirplusAuto as another flag, there could be reason that's why kernel exposed it as another flag.

fuseutil/dirent.go
```diff
@@ -50,6 +51,19 @@ type Dirent struct {
...
+// A struct representing an entry along with its attributes within a directory file, describing a child.
+// See notes on fuseops.ReadDirPlusOp and on WriteDirentPlus for details.
+type DirentPlus struct {
+	// The basic directory entry information (offset, inode, name, type).
+	Dirent Dirent
+
+	// Detailed information about the child inode, including its attributes
+	// (like size, mode, timestamps) and cache expiration times for both the
+	// attributes and the name lookup. This corresponds to the data
+	// consumed by the kernel to populate its cache (fuse_entry_out).
```
raj-prince, line 63:
```suggestion
	// Detailed information about the child inode, including its attributes
	// (like size, mode, timestamps) and cache expiration times for both the
	// attributes and the name lookup.
```

fuseutil/dirent.go
```diff
@@ -103,3 +117,128 @@ func WriteDirent(buf []byte, d Dirent) (n int) {
 
 	return n
 }
+
+// Write the supplied directory entry with attributes into the given buffer in the format
+// expected in fuseops.ReadFileOp.Data, returning the number of bytes written.
```
raj-prince, line 122: You mean?

```suggestion
// expected in fuseops.ReadDirPlusOp.Dst, returning the number of bytes written.
```

fuseutil/dirent.go
```diff
@@ -103,3 +117,128 @@ func WriteDirent(buf []byte, d Dirent) (n int) {
...
+		name    [0]byte
+	}
+
+	type fuse_direntplus struct {
+		entry_out fuse_entry_out
+		dirent    fuse_dirent
+	}
+
+	const direntPlusAlignment = 8
+	const direntPlusHeaderSize = 8 + 8 + 8 + 8 + 4 + 4 + //fuse_entry_out without attributes
```
raj-prince, line 172: Can we do something like this?

```go
const (
    // size of fuse_entry_out (without attributes)
    fuseEntryOutSize = unsafe.Sizeof(fuse.EntryOut{})

    // size of fuse_attr
    fuseAttrSize = unsafe.Sizeof(fuse.Attr{})

    // size of fuse_dirent
    fuseDirentSize = unsafe.Sizeof(fuse.Dirent{})
)

const direntPlusHeaderSize = fuseEntryOutSize + fuseAttrSize + fuseDirentSize
```

fuseutil/dirent.go
```diff
@@ -103,3 +117,128 @@ func WriteDirent(buf []byte, d Dirent) (n int) {
...
+	// for the next entry.
+	var padLen int
+	if len(d.Dirent.Name)%direntPlusAlignment != 0 {
+		padLen = direntPlusAlignment - (len(d.Dirent.Name) % direntPlusAlignment)
+	}
+
+	// Do we have enough room?
+	totalLen := direntPlusHeaderSize + len(d.Dirent.Name) + padLen
+	if totalLen > len(buf) {
+		return n
```
raj-prince, line 186:
```suggestion
		return 0
```

fuseutil/dirent.go
```diff
@@ -103,3 +117,136 @@ func WriteDirent(buf []byte, d Dirent) (n int) {
...
+		nodeid           uint64
+		generation       uint64
+		entry_valid      uint64
+		attr_valid       uint64
+		entry_valid_nsec uint32
+		attr_valid_nsec  uint32
+		attr             fuse_attr
+	}
+
+	type fuse_dirent struct {
```
stapelberg, line 154: Let’s move the fuse_dirent type outside of the function so that we don’t need to duplicate it between WriteDirent and WriteDirentPlus.

fuseutil/dirent.go
```diff
@@ -103,3 +117,136 @@ func WriteDirent(buf []byte, d Dirent) (n int) {
...
+	// according to FUSE_DIRENT_ALIGN (https://tinyurl.com/3m3ewu7h), which
+	// dictates 8-byte alignment.
+	type fuse_direntplus struct {
+		entry_out fuse_entry_out
+		dirent    fuse_dirent
+	}
+
+	const direntPlusAlignment = 8
+
+	//size of fuse_attr
```
stapelberg, line 173: add space after //

fuseutil/dirent.go
```diff
@@ -103,3 +117,136 @@ func WriteDirent(buf []byte, d Dirent) (n int) {
...
+
+	// size of fuse_dirent
+	const fuseDirentSize = 8 + 8 + 4 + 4
+
+	const direntPlusHeaderSize = fuseAttrSize + fuseEntryOutSize + fuseDirentSize
+
+	// Compute the number of bytes of padding we'll need to maintain alignment
+	// for the next entry.
+	var padLen int
+	if len(d.Dirent.Name)%direntPlusAlignment != 0 {
```
stapelberg, line 187: Don’t compute the same expression twice, instead assign it to a variable in the if conditional:

```go
if pad := len(d.Dirent.Name)%direntPlusAlignment; pad != 0 {
	padLen = direntPlusAlignment - pad
}
```

fuseutil/dirent.go
```diff
@@ -103,3 +117,136 @@ func WriteDirent(buf []byte, d Dirent) (n int) {
...
+		padLen = direntPlusAlignment - (len(d.Dirent.Name) % direntPlusAlignment)
+	}
+
+	// Do we have enough room?
+	totalLen := int(direntPlusHeaderSize) + len(d.Dirent.Name) + padLen
+	if totalLen > len(buf) {
+		return 0
+	}
+
+	EntryValid, EntryValidNsec := fuse.ConvertExpirationTime(d.Entry.EntryExpiration)
```
stapelberg, line 197: It’s unusual for variables to have uppercase names in Go, because uppercase means exported. Here and below, use lowercase consistently.

fuseutil/dirent.go
```diff
@@ -103,3 +117,136 @@ func WriteDirent(buf []byte, d Dirent) (n int) {
...
+			nodeid:           uint64(d.Entry.Child),
+			generation:       uint64(d.Entry.Generation),
+			entry_valid:      EntryValid,
+			attr_valid:       AttrValid,
+			entry_valid_nsec: EntryValidNsec,
+			attr_valid_nsec:  AttrValidNsec,
+			attr: fuse_attr{
+				ino:       uint64(d.Entry.Child),
+				size:      d.Entry.Attributes.Size,
+				blocks:    (d.Entry.Attributes.Size + 512 - 1) / 512, // round up to the nearest 512 boundary (In POSIX a "block" is a unit of 512 bytes)
```
stapelberg, line 217: This line is too long. Move the comment from the end of the line to above the line.

fuseutil/dirent.go
```diff
@@ -103,3 +117,136 @@ func WriteDirent(buf []byte, d Dirent) (n int) {
 
 	return n
 }
+
+// Write the supplied directory entry with attributes into the given buffer in the format
+// expected in fuseops.ReadDirPlusOp.Dst returning the number of bytes written.
+// Return zero if the entry would not fit.
+func WriteDirentPlus(buf []byte, d DirentPlus) (n int) {
+	type fuse_attr struct {
```
stapelberg, line 125: This struct seems to duplicate `fusekernel.Attr`? Could we use `fusekernel.Attr` instead?)

mount_config.go
```diff
@@ -205,6 +205,12 @@ type MountConfig struct {
 	// without O_TRUNC, followed by a SetInodeAttributes op with the target size set to 0.
 	// Ref: https://github.com/torvalds/linux/commit/6ff958edbf39c014eb06b65ad25b736be08c4e63
 	EnableAtomicTrunc bool
+
+	// Flag to enable the ReadDirPlus capability.
+	// When enabled, more comprehensive directory information can be retrieved,
```
stapelberg, line 210: This description is not really getting to the point of the feature. How about:

```go
// Flag to tell the kernel we support ReadDirPlus, which optimizes performance
// by returning not just the directory entries (like ReadDir), but also their inode
// attributes, thereby saving one extra Lookup request per directory entry.
```

fusetesting/readdirplus.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
raj-prince, line 1: nit: please change the year -> 2025

fuseutil/dirent.go
aditimittal2003: My apologies, I missed that. I've now added the spaces after // and committed the changes.

fusetesting/readdirplus.go
```diff
@@ -0,0 +1,171 @@
...
+		name: fileName,
+		size: fileSize,
+		mode: fileMode,
+	}, nil
+}
+
+// ReadDirPlusPicky executes long listing command on the given directory and parses its output
+// into a sorted list of os.FileInfo objects.
+func ReadDirPlusPicky(dirname string) (entries []os.FileInfo, err error) {
+	cmd := exec.Command("ls", "-l", dirname)
```
stapelberg, line 122: Why do we need to use `ls -l`, i.e. an external subprocess, in this test? 

The fusetesting/readdir.go equivalent (ReadDirPicky) uses os.File.Readdirnames and os.Lstat directly, which I think is more elegant because we can get rid of all the text parsing.

stapelberg (changes requested):

fusetesting/readdirplus.go
aditimittal2003: There does not exist any sys call for Readdirplus and ReadDirPicky does not invoke ReadDirPlus. So, we need to use ```ls -l``` to call ReadDirPlus.

stapelberg: There must be a way to trigger this behavior from Go. Which system calls does `ls -l` do in this scenario? Check using `strace -fvy -s2048 ls -l`

We can do any system call from Go, so I don’t think we need to rely on calling `ls` here.

stapelberg (approved):

PR #179, opened 2025-07-14
Returning custom error while unmounting /dev/fd/N kind of mount point

raj-prince:
**Issue:**
- Misleading error logs while unmount GCSFuse in GKE world, where mount is created externally and passed via GCSfuse in /dev/fd/N format.
```text
"Failed to unmount in response to SIGTERM: exec: \"fusermount\": executable file not found in $PATH
```
- This happens, as jacobsa/fuse library tries to unmount using fusermount binary, which is not present in GKE container environment and leads to failure. 

**Proposal:**
- Return a custom error during fuse.Unmount() so that file-system implementation can handle appropriately.

**Testing:**
Working as expected.

### DIFF
--- a/unmount_linux_test.go
+++ b/unmount_linux_test.go
@@ -6,22 +6,18 @@ import (
 )
 
 func Test_umountExpectCustomError(t *testing.T) {
-	dir := "/dev/fd/42"
 	t.Setenv("PATH", "") // Clear PATH to fail unmount with fusermount is not found
 
-	err := unmount(dir)
-
+	err := unmount("/dev/fd/42")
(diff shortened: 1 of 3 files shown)


unmount_linux_test.go
```diff
@@ -0,0 +1,31 @@
+package fuse
+
+import (
+	"errors"
+	"testing"
+)
+
+func Test_umountExpectCustomError(t *testing.T) {
+	dir := "/dev/fd/42"
+	fuserunmountMock = func(string) error {
```
stapelberg, line 10: Instead of mocking the function, would it be possible to use `t.Setenv("PATH", "")` here and thereby make `exec.LookPath` return an error? That way, we don’t need to change the non-test code only for testing.

PR #182, opened 2025-08-27
fix: false success in unmount flow

raj-prince:
Unmount flow started returning false success during the unmount after [this](https://github.com/jacobsa/fuse/pull/179).

### DIFF
--- a/unmount_linux_test.go
+++ b/unmount_linux_test.go
@@ -18,7 +18,7 @@ func Test_umountNoCustomError(t *testing.T) {
 	t.Setenv("PATH", "") // Clear PATH to fail unmount with fusermount is not found
 
 	err := unmount("/dev")
-	if err != nil && errors.Is(err, ErrExternallyManagedMountPoint) {
-		t.Errorf("Not expected custom error.")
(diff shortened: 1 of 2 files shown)


unmount_linux_test.go
```diff
@@ -18,7 +18,7 @@ func Test_umountNoCustomError(t *testing.T) {
 	t.Setenv("PATH", "") // Clear PATH to fail unmount with fusermount is not found
 
 	err := unmount("/dev")
-	if err != nil && errors.Is(err, ErrExternallyManagedMountPoint) {
-		t.Errorf("Not expected custom error.")
+	if err == nil || errors.Is(err, ErrExternallyManagedMountPoint) {
+		t.Errorf("Expected error but not the custom error.")
```
gargnitingoogle, line 22: The error message is misleading in case `err == nil`. It should be something like.

```go
if err == nil {
  t.Errorf("Expected error, but got none.")
} else if errors.Is(err, ErrExternallyManagedMountPoint) {
  t.Errorf("Got custom error which was unexpected.")
}
```

raj-prince: Added separate assertion.

PR #188, opened 2026-01-21
Support kill-priv flags in init handshake

sbauersfeld:
By default, the kernel will send a GetXAttrs("security.capabilities") request to the FUSE daemon before every Write operation. It does this because the kernel may need to clear setuid/setgid bits and security capabilities when a file is written. This incurs a performance cost because of the extra round trips between kernel and user space when calling GetXAttrs.

Some filesystems do not even support security capabilities. Setting the HandleKillPriv flags allows those filesystems to avoid the extra GetXAttrs calls.

Another option is to return ENOSYS for the GetXAttr call, but that also means that the filesystem cannot support GetXAttr calls for any attribute at all.

This change allows filesystems to make use of the HandleKillPriv flags.

See comments here with more explanation of how HandleKillPriv works: https://github.com/jacobsa/fuse/pull/188/changes#diff-a3f464fe0022dffde22768b97e434466a904c3a6955b11d2f19f85800fca12f7R249

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -150,6 +150,11 @@ func (c *Connection) Init() error {
 		c.protocol = initOp.Kernel
 	}
 
+	if c.debugLogger != nil {
+		c.debugLogger.Printf("Kernel protocol version major: %v, minor: %v\n", initOp.Kernel.Major, initOp.Kernel.Minor)
+		c.debugLogger.Printf("Protocol version major: %v, minor: %v\n", c.protocol.Major, c.protocol.Minor)
+	}
+
 	cacheSymlinks := initOp.Flags&fusekernel.InitCacheSymlinks > 0
 	noOpenSupport := initOp.Flags&fusekernel.InitNoOpenSupport > 0
 	noOpendirSupport := initOp.Flags&fusekernel.InitNoOpendirSupport > 0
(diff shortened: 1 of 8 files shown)


sbauersfeld:
fyi @jacobsa, requesting to add support for HandleKillPriv and HandleKillPrivV2. The implementation is fairly straightforward but the exact behavior of kernel interactions with the fuse daemon when HandleKillPrivV2 is enabled is a bit strange. I tried to add sufficient test coverage and documentation to cover the relevant interactions. Please take a look when you have some free time!

sbauersfeld:
friendly bump on this @stapelberg :)

sbauersfeld:
Hey @stapelberg @geertj , just checking back in here

sbauersfeld:
@stapelberg , is there any possibility I could become a maintainer for this project and help review and merge PRs like this one? I am also interested in adding support for fuse passthrough. I would prefer to contribute to an existing open source project like this one rather than create a fork. Thanks, Scott

sbauersfeld:
Hi @kislaykishore I see you made the most recent contribution for this project, do you have any advice as to how I can get this change reviewed / merged?

sbauersfeld:
checking back in on this @stapelberg

PR #194, opened 2026-04-29
Enable Configurable Maximum Buffer Size for FUSE Messages

abhishek10004:
This PR removes the hardcoded 1 MiB limit for FUSE message buffers in jacobsa/fuse and makes it configurable via MountConfig. This is necessary to support larger kernel requests (e.g., 16 MiB) without triggering buffer overflows that result in 0-byte reads and premature EOFs. (Also, we're still keep the minimum buffer size value to 1MiB which is still the default value)

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -80,6 +80,8 @@ type Connection struct {
 	// Freelists, serviced by freelists.go.
 	inMessages  freelist.Freelist // GUARDED_BY(mu)
 	outMessages freelist.Freelist // GUARDED_BY(mu)
+
+	inMessageSize int
 }
 
 // State that is maintained for each in-flight op. This is stuffed into the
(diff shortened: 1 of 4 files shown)


connection.go
```diff
@@ -121,6 +123,12 @@ func newConnection(
 		cancelFuncs: make(map[uint64]func()),
 	}
 
+	maxPayload := max(buffer.MaxReadSize, buffer.MaxWriteSize)
+	if cfg.MaxMessageSize > 0 {
+		maxPayload = max(maxPayload, int(cfg.MaxMessageSize))
+	}
+	c.inMessageSize = maxPayload + buffer.GetPageSize()
```
vadlakondaswetha, line 130: should we add this pageSize() inside in_message.go constructor? that way we can avoid subtracting for maxWrite. 

you can also rename this parameter to maxPayload

vadlakondaswetha (approved):

PR #195, opened 2026-06-02
Expose FUSE_ASYNC_DIO in MountConfig

geertj:
When FUSE handles a read or write operation that is larger than the maximum FUSE message size it has to break up the request into multiple smaller ones. When this flag is set, FUSE will send these smaller requests in parallel to the FUSE server, which can then start the IO for all of them at the same time. This can significanlty increase throughput for remote file systems such as GCSFuse.

On current Linux kernels, the FUSE_ASYNC_DIO works for async IO (io_uring and libaio), but not for the traditional Posix read() and write() operations.


abhishek10004 (approved):
LGTM

PR #198, opened 2026-06-22
Support Configurable Large Messages (> 1 MiB) via Zero-Copy Vectored I/O

abhishek10004:
## Overview
Previously, this library had a hardcoded FUSE message buffer size of `1 MiB + pageSize` (corresponding to the standard 1 MiB FUSE payload limit). Because of this hardcoded limit, **large reads and writes (> 1 MiB) were not supported at all**. 

This PR adds support for large I/O operations by introducing a configurable `MaxMessageSize` in `MountConfig`, allowing daemons to read and write messages larger than 1 MiB. 

To prevent the severe performance regressions, memory fragmentation, and garbage collection (GC) pressure that would arise from allocating giant contiguous buffers (e.g., 4 MiB, 8 MiB, or 16 MiB) on the heap for every request, **this support is implemented via Vectored I/O**:
1. **Large Reads (> 1 MiB)** are read from the FUSE device directly into non-contiguous, block-pooled buffers (1 MiB blocks) via the `readv` system call. The filesystem can then write the read payload directly into these blocks via `ReadFileOp.DstBufs` (**Zero-Copy Vectored Reads**).
2. **Large Writes (> 1 MiB)** bypass copying the incoming payload into a single contiguous slice, instead exposing the raw non-contiguous block slices directly to the filesystem via `WriteFileOp.DataBlocks` (**Zero-Copy Vectored Writes**).

---

## Key Benefits
* **Support for Large I/O (> 1 MiB):** Enables high-throughput FUSE operations by removing the hardcoded 1 MiB message size ceiling.
* **Zero-Copy & Low GC Pressure:** Avoids massive heap allocations and contiguous memory copies for large transfers by leveraging a thread-safe, block-pooled allocator (`BlockPool1M` and `BlockPool1MPlusPage`) and the `readv` system call.
* **Backward Compatibility:** Preserves contiguous buffers (`Dst` and `Data`) for reads/writes under 1 MiB or when vectored I/O is disabled, ensuring existing filesystem implementations continue to work out-of-the-box.

---

### 1. Enable Setting FUSE Buffer Sizes Dynamically
**Commit:** `ddc386c`
* **Purpose:** Lays the foundation for configurable message sizes by allowing the buffer size to be set dynamically prior to mounting, rather than relying on a hardcoded global limit.
* **Key Changes:**
  * Added `MaxMessageSize uint32` to [mount_config.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/mount_config.go).
  * Dynamically calculates `c.inMessageSize` in [connection.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/connection.go) based on `MaxMessageSize` (or defaults to the maximum of `MaxReadSize`/`MaxWriteSize` + 1 page).
  * Configures the FUSE protocol initialization (`Init` call) to announce this dynamically calculated size as `MaxWrite` and `MaxPages` to the FUSE kernel driver.
  * Refactored `InMessage` in [internal/buffer/in_message.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/internal/buffer/in_message.go) to accept a dynamic allocation size instead of using a global static `bufSize`.

### 2. Refactor Error Handling using standard `errors.Is`
**Commit:** `c263028`
* **Purpose:** Cleans up error-handling logic when reading FUSE requests to make it more robust.
* **Key Changes:**
  * Refactored `Connection.readMessage()` in [connection.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/connection.go) to use standard `errors.Is(err, syscall.ENODEV)` and `errors.Is(err, syscall.EINTR)` checks instead of type-casting to `*os.PathError` and inspecting inner fields. This ensures compatibility in case errors are wrapped.

### 3. Implement Zero-Copy Vectored Reads Support
**Commit:** `72acb69`
* **Purpose:** Implements the infrastructure for reading FUSE requests from the device directly into non-contiguous blocks, and exposing these blocks to the filesystem via `ReadFileOp.DstBufs` to support large reads (> 1 MiB) without heap thrashing.
* **Key Changes:**
  * **Block-Pool Allocator:**
    * Defined thread-safe pools `BlockPool1M` (1 MiB blocks) and `BlockPool1MPlusPage` (1 MiB + hardware page size) in [internal/buffer/in_message.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/internal/buffer/in_message.go) to recycle buffers and avoid heap thrashing.
    * Refactored `InMessage` to allocate non-contiguous blocks (`blocks [][]byte`) rather than a single contiguous slice. Block 0 is always sized `1 MiB + pageSize` (holding headers and small payloads), and additional 1 MiB blocks are allocated to satisfy larger message limits.
  * **Zero-Copy Syscall (`readv`):**
    * Added [internal/buffer/readv.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/internal/buffer/readv.go) which implements a wrapper around the `SYS_READV` syscall, converting block slices to `unix.Iovec` pointers to perform a single-system-call read into multiple non-contiguous memory segments.
    * Retained backward compatibility on macOS (`FuseT`) by implementing a contiguous fallback pool (`fuseTContiguousPool`).
  * **Vectored Read API:**
    * Added `EnableVectoredReads` to `MountConfig`.
    * Added `DstBufs [][]byte` to [fuseops/ops.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/fuseops/ops.go#L711-L731). When enabled and the read size is larger than block 0, `Dst` is set to `nil` and `DstBufs` is populated with the block-sliced buffers, allowing the filesystem to write read payloads directly into the FUSE message blocks.
  * **Testing:**
    * Added over 760 lines of comprehensive unit tests in [internal/buffer/in_message_test.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/internal/buffer/in_message_test.go) covering block allocations, boundary-spanning data consumption, vector slicing, and pool returns.

### 4. Implement Zero-Copy Vectored Writes Support & MemFS Optimization
**Commit:** `8005ed3`
* **Purpose:** Adds zero-copy support for incoming large FUSE write requests, bypassing contiguous payload reconstruction, and optimizes the `memfs` memory filesystem.
* **Key Changes:**
  * **Vectored Writes API:**
    * Added `EnableVectoredWrites` to `MountConfig`.
    * Added `DataBlocks [][]byte` and a `TotalSize() int` helper method to [fuseops/ops.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/fuseops/ops.go#L809).
    * When enabled, `convertInMessage` slices the write payload directly into `DataBlocks` (using `ConsumeVector`), completely avoiding copying the payload into a single contiguous slice.
  * **MemFS Optimization:**
    * Added `WriteBlocksAt(blocks [][]byte, off int64)` to [samples/memfs/inode.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/samples/memfs/inode.go#L375-L412) to copy block-by-block directly into the inode's storage slice.
    * Updated `WriteFile` in [samples/memfs/memfs.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/samples/memfs/memfs.go#L694) to leverage `WriteBlocksAt` when `DataBlocks` is populated.
  * **Wirelog, Debug, and Test Cleanups:**
    * Simplified write size calculations in `debug.go` and `wirelog.go` to use the new `WriteFileOp.TotalSize()` helper method.
    * Added integration tests verifying `VectoredWritesTest` in [samples/memfs/memfs_test.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/samples/memfs/memfs_test.go#L2132).
    * Fixed out-of-cache benchmarks in [internal/buffer/out_message_test.go](file:///usr/local/google/home/abhishekmgupta/jacobsa_fuse/internal/buffer/out_message_test.go#L276) by allocating larger arrays (80 MiB) on the heap to successfully defeat the CPU cache.

---

## Architectural Design: Why Vectored I/O?

To support message sizes larger than 1 MiB, allocating contiguous buffers dynamically (e.g., a single 8 MiB buffer for an 8 MiB read/write) is highly inefficient due to severe GC pressure and heap fragmentation. 

Instead, the library now implements a **non-contiguous, block-based architecture**:
```
           +-----------------------+      +-------------------+
InMessage  | Block 0 (1MB + page)  | ---> | Block 1 (1MB)     | ---> ...
           +-----------------------+      +-------------------+
           | Headers | Small data  |      | Large data blocks |
           +-----------------------+      +-------------------+
```

### 1. Zero-Copy Reads
When a FUSE read request is received:
* The daemon uses the `readv` syscall on Linux to read data directly from the `/dev/fuse` descriptor into the pooled blocks, avoiding any kernel-to-user memory copy.
* If the read size is larger than Block 0, the library populates `ReadFileOp.DstBufs` with these blocks.
* The filesystem writes the data directly into `DstBufs`, requiring zero extra allocations or copy operations.

### 2. Zero-Copy Writes
When a FUSE write request is received:
* The write payload is read into the pooled blocks.
* Instead of allocating a single contiguous buffer and copying all blocks into it, the library returns the non-contiguous slices directly in `WriteFileOp.DataBlocks`.
* Filesystems optimized for vectored writes (such as the updated `memfs` using `WriteBlocksAt`) can consume these blocks directly.

---

## Configuration Reference
Three new fields are introduced in `MountConfig` to manage the large message and vectored I/O behavior:

| Field | Type | Description |
| :--- | :--- | :--- |
| `MaxMessageSize` | `uint32` | Configures the maximum size of FUSE messages the daemon is prepared to read/write. **Setting this larger than 1 MiB enables large reads and writes.** |
| `EnableVectoredReads` | `bool` | If true, large read operations bypass contiguous buffer allocations in `ReadFileOp.Dst` and instead populate `ReadFileOp.DstBufs`. |
| `EnableVectoredWrites` | `bool` | If true, large write operations bypass copying payload blocks into a single contiguous slice in `WriteFileOp.Data` and instead populate `WriteFileOp.DataBlocks`. |

> [!IMPORTANT]
> For `MaxMessageSize` values greater than 1 MiB, enabling both `EnableVectoredReads` and `EnableVectoredWrites` is highly recommended to avoid significant performance regressions due to large heap allocations and contiguous copies.

---

## Testing & Verification
* **Unit Tests:** `internal/buffer/in_message_test.go` verifies all edge cases of multi-block message parsing (consuming across block boundaries, shrinking, and block recycling).
* **Integration Tests:** `samples/memfs/memfs_test.go` contains test cases ensuring that MemFS correctly handles vectored write operations when `EnableVectoredWrites` is enabled.
* **Benchmarks:** Benchmarks in `internal/buffer/out_message_test.go` have been updated and verified to measure reset, growth, and shrink performance accurately.
* **All tests pass successfully.**

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -80,6 +80,8 @@ type Connection struct {
 	// Freelists, serviced by freelists.go.
 	inMessages  freelist.Freelist // GUARDED_BY(mu)
 	outMessages freelist.Freelist // GUARDED_BY(mu)
+
+	inMessageSize int
 }
 
 // State that is maintained for each in-flight op. This is stuffed into the
@@ -120,6 +122,12 @@ func newConnection(
 		dev:         dev,
 		cancelFuncs: make(map[uint64]func()),
 	}
+
+	maxPayload := max(buffer.MaxReadSize, buffer.MaxWriteSize)
+	if cfg.MaxMessageSize > 0 {
+		maxPayload = max(maxPayload, int(cfg.MaxMessageSize))
+	}
+	c.inMessageSize = maxPayload + buffer.GetPageSize()
 
 	// Initialize.
 	if err := c.Init(); err != nil {
@@ -172,7 +180,9 @@ func (c *Connection) Init() error {
 	// Respond to the init op.
 	initOp.Library = c.protocol
 	initOp.MaxReadahead = maxReadahead
-	initOp.MaxWrite = buffer.MaxWriteSize
+	
+	maxPayload := c.inMessageSize - buffer.GetPageSize()
+	initOp.MaxWrite = uint32(maxPayload)
 
 	initOp.Flags = 0
 
@@ -190,7 +200,6 @@ func (c *Connection) Init() error {
 	// payload. It applies to both requests and replies, and does not include
 	// the extra 1 page for the FUSE header and the "args" struct. We set it to
 	// the max of our message in/out payload sizes.
-	maxPayload := max(buffer.MaxReadSize, buffer.MaxWriteSize)
 	initOp.MaxPages = uint16(maxPayload / buffer.GetPageSize())
 
 	// Enable writeback caching if the user hasn't asked us not to.
--- a/connection.go
+++ b/connection.go
@@ -16,6 +16,7 @@ package fuse
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -398,17 +399,12 @@ func (c *Connection) readMessage() (*buffer.InMessage, error) {
 		//  *  EINTR means we should try again. (This seems to happen often on
 		//     OS X, cf. http://golang.org/issue/11180)
 		//
-		if pe, ok := err.(*os.PathError); ok {
-			switch pe.Err {
-			case syscall.ENODEV:
-				err = io.EOF
-
-			case syscall.EINTR:
-				err = nil
-				continue
-			}
+		if errors.Is(err, syscall.ENODEV) {
+			err = io.EOF
+		} else if errors.Is(err, syscall.EINTR) {
+			err = nil
+			continue
 		}
-
 		if err != nil {
 			c.putInMessage(m)
 			return nil, err
--- a/connection.go
+++ b/connection.go
@@ -181,7 +181,7 @@ func (c *Connection) Init() error {
 	// Respond to the init op.
 	initOp.Library = c.protocol
 	initOp.MaxReadahead = maxReadahead
-	
+
 	maxPayload := c.inMessageSize - buffer.GetPageSize()
 	initOp.MaxWrite = uint32(maxPayload)
 
@@ -386,6 +386,7 @@ func (c *Connection) handleInterrupt(fuseID uint64) {
 func (c *Connection) readMessage() (*buffer.InMessage, error) {
 	// Allocate a message.
 	m := c.getInMessage()
+	m.AllocBlocks(c.inMessageSize)
 
 	// Loop past transient errors.
 	for {
@@ -405,6 +406,7 @@ func (c *Connection) readMessage() (*buffer.InMessage, error) {
 			err = nil
 			continue
 		}
+
 		if err != nil {
 			c.putInMessage(m)
 			return nil, err
--- a/conversions.go
+++ b/conversions.go
@@ -145,13 +145,14 @@ func convertInMessage(
 		}
 
 		entries := make([]fuseops.BatchForgetEntry, 0, in.Count)
-		for i := uint32(0); i < in.Count; i++ {
-			type entry fusekernel.BatchForgetEntryIn
-			ein := (*entry)(inMsg.Consume(unsafe.Sizeof(entry{})))
-			if ein == nil {
-				return nil, errors.New("Corrupt OpBatchForget")
-			}
+		entrySize := unsafe.Sizeof(fusekernel.BatchForgetEntryIn{})
+		buf := inMsg.ConsumeBytes(uintptr(in.Count) * entrySize)
+		if len(buf) < int(in.Count*uint32(entrySize)) {
+			return nil, errors.New("Corrupt OpBatchForget")
+		}
 
+		for i := uint32(0); i < in.Count; i++ {
+			ein := (*fusekernel.BatchForgetEntryIn)(unsafe.Pointer(&buf[uintptr(i)*entrySize]))
 			entries = append(entries, fuseops.BatchForgetEntry{
 				Inode: fuseops.InodeID(ein.Inode),
 				N:     ein.Nlookup,
@@ -396,7 +397,11 @@ func convertInMessage(
 			},
 		}
 		// Use part of the incoming message storage as the read buffer.
-		to.Dst = inMsg.GetFree(int(in.Size))
+		if config.EnableVectoredReads && int(in.Size) > buffer.MiBPlusPageSize {
+			to.DstBufs = inMsg.GetFreeVector(int(in.Size))
+		} else {
+			to.Dst = inMsg.GetFree(int(in.Size))
+		}
 		o = to
 
 	case fusekernel.OpReaddir:
@@ -933,6 +938,8 @@ func (c *Connection) kernelResponseForOp(
 	case *fuseops.ReadFileOp:
 		if o.Data != nil {
 			m.Append(o.Data...)
+		} else if o.DstBufs != nil {
+			m.Append(o.DstBufs...)
 		} else {
 			m.Append(o.Dst)
 		}
--- a/conversions.go
+++ b/conversions.go
@@ -503,16 +503,31 @@ func convertInMessage(
 			return nil, errors.New("Corrupt OpWrite")
 		}
 
-		buf := inMsg.ConsumeBytes(inMsg.Len())
-		if len(buf) < int(in.Size) {
-			return nil, errors.New("Corrupt OpWrite")
+		var buf []byte
+		var dataBlocks [][]byte
+
+		if config.EnableVectoredWrites && inMsg.Len() > uintptr(buffer.MiBPlusPageSize) {
+			dataBlocks = inMsg.ConsumeVector(inMsg.Len())
+			var totalLen int
+			for _, b := range dataBlocks {
+				totalLen += len(b)
+			}
+			if totalLen < int(in.Size) {
+				return nil, errors.New("Corrupt OpWrite")
+			}
+		} else {
+			buf = inMsg.ConsumeBytes(inMsg.Len())
+			if len(buf) < int(in.Size) {
+				return nil, errors.New("Corrupt OpWrite")
+			}
 		}
 
 		o = &fuseops.WriteFileOp{
-			Inode:  fuseops.InodeID(inMsg.Header().Nodeid),
-			Handle: fuseops.HandleID(in.Fh),
-			Data:   buf,
-			Offset: int64(in.Offset),
+			Inode:      fuseops.InodeID(inMsg.Header().Nodeid),
+			Handle:     fuseops.HandleID(in.Fh),
+			Data:       buf,
+			DataBlocks: dataBlocks,
+			Offset:     int64(in.Offset),
 			OpContext: fuseops.OpContext{
 				FuseID: inMsg.Header().Unique,
 				Pid:    inMsg.Header().Pid,
@@ -947,7 +962,7 @@ func (c *Connection) kernelResponseForOp(
 
 	case *fuseops.WriteFileOp:
 		out := (*fusekernel.WriteOut)(m.Grow(int(unsafe.Sizeof(fusekernel.WriteOut{}))))
-		out.Size = uint32(len(o.Data))
+		out.Size = uint32(o.TotalSize())
 
 	case *fuseops.SyncFileOp:
 		// Empty response
--- a/internal/buffer/in_message.go
+++ b/internal/buffer/in_message.go
@@ -28,13 +28,8 @@ import (
 // this.
 var pageSize int
 
-// We size the buffer to have enough room for a fuse request plus data
-// associated with a write request.
-var bufSize int
-
 func init() {
 	pageSize = syscall.Getpagesize()
-	bufSize = pageSize + MaxWriteSize
 }
 
 // Return the hardware page size. Note that this is not always 4KiB! Notably
@@ -53,9 +48,9 @@ type InMessage struct {
 }
 
 // NewInMessage creates a new InMessage with its storage initialized.
-func NewInMessage() *InMessage {
+func NewInMessage(size int) *InMessage {
 	return &InMessage{
-		storage: make([]byte, bufSize),
+		storage: make([]byte, size),
 	}
 }
 
--- a/internal/buffer/in_message.go
+++ b/internal/buffer/in_message.go
@@ -22,14 +22,26 @@ import (
 	"unsafe"
 
 	"github.com/jacobsa/fuse/internal/fusekernel"
+	"golang.org/x/sys/unix"
 )
 
 // All requests read from the kernel, without data, are shorter than
 // this.
 var pageSize int
 
+// Constants for buffer/message sizes.
+const (
+	// MiB is 1 MiB in bytes.
+	MiB = 1024 * 1024
+)
+
+// MiBPlusPageSize is 1 MiB + hardware page size. Since pageSize is determined
+// at runtime, this is a variable rather than a compile-time constant.
+var MiBPlusPageSize int
+
 func init() {
-	pageSize = syscall.Getpagesize()
+	pageSize = unix.Getpagesize()
+	MiBPlusPageSize = MiB + pageSize
 }
 
 // Return the hardware page size. Note that this is not always 4KiB! Notably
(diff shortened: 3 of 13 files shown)


internal/buffer/in_message.go
```diff
@@ -43,52 +50,213 @@ func GetPageSize() int {
...
+		p.list = append(p.list, buf)
+		p.mu.Unlock()
+		return
+	}
+	p.mu.Unlock()
+	p.overflow.Put(buf)
+}
+
+var BlockPool1M = newBlockPool(48, func() []byte {
+	return make([]byte, MiB)
```
abhishek10004, line 98: I've not used mmap here because currently I'm overflowing to a syncPool and hence if there is a lot of parallelism, we would not be paying the allocation penalty again and again. 

In case we take over the memory allocation using mmap and bypass the go runtime, then there would be 2 options: 
a) fixed size buffer pool but that would mean constant allocation/deallocation in case parallelism is higher than the configured limits 
b) dynamic pool that keeps growing/shrinking but this would be a slightly larger change & would need more testing. 
Hence, I've parked it for later, either as a separate commit or a new change.

connection.go
```diff
@@ -168,12 +177,13 @@ func (c *Connection) Init() error {
...
 	noOpenSupport := initOp.Flags&fusekernel.InitNoOpenSupport > 0
 	noOpendirSupport := initOp.Flags&fusekernel.InitNoOpendirSupport > 0
-	asyncDIO := initOp.Flags&fusekernel.InitAsyncDIO > 0
 
 	// Respond to the init op.
 	initOp.Library = c.protocol
 	initOp.MaxReadahead = maxReadahead
-	initOp.MaxWrite = buffer.MaxWriteSize
+
+	maxPayload := c.inMessageSize - buffer.GetPageSize()
```
vadlakondaswetha, line 185: do we anticipate different sizes for read and write. if not can we have just one variable which tells size of the request for both reads and writes.

internal/buffer/in_message.go
```diff
@@ -43,52 +50,213 @@ func GetPageSize() int {
...
-	size      int
+	blocks         [][]byte
+	size           int
+	consumed       int
+	iovecs         []unix.Iovec
+	borrowedBlocks [][]byte
+}
+
+// NewInMessage creates a new InMessage.
+func NewInMessage(size int) *InMessage {
```
vadlakondaswetha, line 117: why are you taking a size parameter if its not used.

connection.go
```diff
@@ -394,15 +400,11 @@ func (c *Connection) readMessage() (*buffer.InMessage, error) {
...
-		if pe, ok := err.(*os.PathError); ok {
-			switch pe.Err {
-			case syscall.ENODEV:
-				err = io.EOF
-
-			case syscall.EINTR:
-				err = nil
-				continue
-			}
+		if errors.Is(err, syscall.ENODEV) {
```
vadlakondaswetha, line 403: we are removing a typecasting here? Are these changes intentiontal? If yes, how did they work earlier?

internal/buffer/in_message.go
```diff
@@ -43,52 +50,213 @@ func GetPageSize() int {
...
+// Initialize with the data read by a single call to r.Read or readv. The first call to
 // Consume will consume the bytes directly after the fusekernel.InHeader
 // struct.
 func (m *InMessage) Init(r io.Reader) error {
-
 	var n int
 	var err error
 	if fusekernel.IsPlatformFuseT {
-		n, err = m.ReadSingle(r)
+		if len(m.blocks) == 1 {
```
vadlakondaswetha, line 207: as discussed please remove all changes for MAC and throw not supported exception when messageSize is bigger. Lets not checkin changes which are not reviewed.

internal/buffer/in_message.go
```diff
@@ -43,52 +50,213 @@ func GetPageSize() int {
 	return pageSize
 }
 
+type blockPool struct {
```
vadlakondaswetha, line 53: Please create a seperate file for blockPool

internal/buffer/in_message.go
```diff
@@ -145,16 +335,110 @@ func (m *InMessage) ConsumeBytes(n uintptr) []byte {
...
+		return [][]byte{m.blocks[0][m.size : m.size+n]}
+	}
+
+	var res [][]byte
+	remaining := n
+	for remaining > 0 {
+		block := BlockPool1M.Get()
+		m.borrowedBlocks = append(m.borrowedBlocks, block)
+		allocSize := MiB
+		if remaining < allocSize {
```
vadlakondaswetha, line 437: what happens if you remove this check?

internal/buffer/in_message.go
```diff
@@ -145,16 +335,110 @@ func (m *InMessage) ConsumeBytes(n uintptr) []byte {
...
 		return nil
 	}
-	return m.storage[m.size : m.size+n]
+	if len(m.blocks) > 0 && m.size+n <= len(m.blocks[0]) {
+		return m.blocks[0][m.size : m.size+n]
+	}
+	// Since n doesn't fit in block 0, and block 0 has size 1MB + pageSize,
+	// n is necessarily larger than 1MB (assuming typical small offset like
+	// sizeof(ReadIn)). Thus we always allocate directly on the heap.
+	return make([]byte, n)
```
vadlakondaswetha, line 417: in the existing code, if we dont required buffer, we are returning nil vs here we are creating a new buffer.
Also i didnt understand in what scenarios would it cross 1MB?

conversions.go
```diff
@@ -396,7 +397,11 @@ func convertInMessage(
 			},
 		}
 		// Use part of the incoming message storage as the read buffer.
-		to.Dst = inMsg.GetFree(int(in.Size))
+		if config.EnableVectoredReads && int(in.Size) > buffer.MiBPlusPageSize {
```
vadlakondaswetha, line 400: Brainstorming a bit here. why do we need to support both vectoredReads and non vectoredReads. Can we just pass 2-D array always. It would be a minor change on the GCSFuse side. How big of a change will it be on GCSFuse side? I am guessing we can just pick the first block from the array and pass it downstream when messageSize is 1MB?

conversions.go
```diff
@@ -498,16 +503,31 @@ func convertInMessage(
 			return nil, errors.New("Corrupt OpWrite")
 		}
 
-		buf := inMsg.ConsumeBytes(inMsg.Len())
-		if len(buf) < int(in.Size) {
-			return nil, errors.New("Corrupt OpWrite")
+		var buf []byte
+		var dataBlocks [][]byte
+
+		if config.EnableVectoredWrites && inMsg.Len() > uintptr(buffer.MiBPlusPageSize) {
```
vadlakondaswetha, line 509: by moving everything to vectoredReads/writes we need not do if-else every where. the code becomes much simpler.

vadlakondaswetha: reduces the number of configs too.

internal/buffer/in_message.go
```diff
@@ -145,16 +335,110 @@ func (m *InMessage) ConsumeBytes(n uintptr) []byte {
...
+	if offset+int(n) <= len(m.blocks[blockIdx]) {
+		b := m.blocks[blockIdx][offset : offset+int(n)]
+		m.consumed += int(n)
+		return b
+	}
+
+	// In production, any spanning allocation is larger than 1MB (since block 0
+	// is 1MB + pageSize and fits all normal headers/payloads). Thus we always
+	// allocate directly from the heap.
+	res := make([]byte, n)
```
vadlakondaswetha, line 349: Same as reads? why would we overflow here and not earlier?

vadlakondaswetha (commented):
yet to review the testcases and samples

connection.go
abhishek10004: The old code was the idiomatic way to handle nested error before Errors.Is was released.
When reading from a device or a file, the error returned is typically wrapped in PathError. Hence, we were unwrapping it earlier and then comparing it. With Errors.Is, this is no longer needed.

internal/buffer/in_message.go
abhishek10004: Then the final block won't be truncated to the actual remaining requested size and we'd be passing a buffer which is larger than the requested data.

conversions.go
abhishek10004: I kept support for non-vectored I/O as well so that it doesn't become a breaking change. Otherwise, anyone updating the library would be forced to make changes to their code, even though those changes would mostly be minor.

internal/buffer/in_message.go
abhishek10004: I'm doing a new allocation here since the read request can go beyond 1MB reads and if the user has not enabled vectored reads, than we'd have to pass a single buffer of requested size to the user.

Regarding earlier case, in that scenario we had a 1MiB+page size buffer and the header is less than page size, so the original buffer would always have 1MiB space & hence the condition "n > len(m.storage)-m.size" would never be true.

internal/buffer/in_message.go
abhishek10004: This new allocation will be used when the user has not enabled vectored writes and increased the fuse max pages limit beyond 1M. In that case, we'd have to allocate a single buffer which can contain the whole content to be written.

conversions.go
abhishek10004: yes, it would become easier & cleaner as well but as mentioned above it'd be a breaking change.

geertj:
Hey @abhishek10004. Thanks for putting this together!

I know that I advocated for the approach of using readv() to allow large messages (e.g. 16MB reads which is what we need on GCS Standard), while at the same time not having to make all messages 16MB. However, after reviewing this PR in some detail, I am not very comfortable with the complexity of what we're adding and the size of this change.

I wonder if instead we need to do something simpler: add a MaxPages config option to MountConfig. If set, that's what we pass in Connection::Init. If unset, we keep the current value based on MaxRead/WriteSize. Note that we'd keep MaxWrite as it is right now.

I think the result would be:
- A read() from the socket never returns more than 1MB + 4KB i.e. our current buffer size for inMessage. This is because the only operation that has an unbounded payload is FUSE_WRITE, and that would be clamped by what we set for MaxWrite in FUSE_INIT.
- The higher value of MaxPages would allow FUSE to send us large FUSE_READ operations, which is what we're after. The request handler needs to use ReadFileOp::Data instead of ReadFileOp::Dst. I think that is a fair API contract for when you set a MaxPages value > 1MB. It is also self documenting because a developer can do a len(op.Dst) and notice that it does not contain sufficient space for the requested bytes.

What do you think? Would this work?

abhishek10004:
> Hey @abhishek10004. Thanks for putting this together!
> 
> I know that I advocated for the approach of using readv() to allow large messages (e.g. 16MB reads which is what we need on GCS Standard), while at the same time not having to make all messages 16MB. However, after reviewing this PR in some detail, I am not very comfortable with the complexity of what we're adding and the size of this change.
> 
> I wonder if instead we need to do something simpler: add a MaxPages config option to MountConfig. If set, that's what we pass in Connection::Init. If unset, we keep the current value based on MaxRead/WriteSize. Note that we'd keep MaxWrite as it is right now.
> 
> I think the result would be:
> 
> * A read() from the socket never returns more than 1MB + 4KB i.e. our current buffer size for inMessage. This is because the only operation that has an unbounded payload is FUSE_WRITE, and that would be clamped by what we set for MaxWrite in FUSE_INIT.
> * The higher value of MaxPages would allow FUSE to send us large FUSE_READ operations, which is what we're after. The request handler needs to use ReadFileOp::Data instead of ReadFileOp::Dst. I think that is a fair API contract for when you set a MaxPages value > 1MB. It is also self documenting because a developer can do a len(op.Dst) and notice that it does not contain sufficient space for the requested bytes.
> 
> What do you think? Would this work?

We can set MaxPages & MaxWrite to different values. In this case, while the reads would be limited by MaxPages value (which means we can get read requests larger than 1MB), the write requests would be limited by the minimum of what MaxPages & MaxWrite evaluates to. Hence, if we keep setting MaxWrite to 1M, we'd not need to use readv and vectored io at all. This would significantly reduce the amount of change. 

Only issue would be that in case of writes, we'd not get larger requests but that would be okay because we still buffer the writes locally (in streaming writes, we buffer in the in-memory block till it's full whereas in staging writes we buffer full writes on disk). Hence, smaller requests would not make that much of difference as the major bottleneck in writes would actually be the upload time.

geertj:
> Only issue would be that in case of writes, we'd not get larger requests but that would be okay because we still buffer the writes locally (in streaming writes, we buffer in the in-memory block till it's full whereas in staging writes we buffer full writes on disk). Hence, smaller requests would not make that much of difference as the major bottleneck in writes would actually be the upload time.

Exactly. Writes will always have to be buffered /somewhere/ (in an actual buffer, or in a network stream) due to the semantics of object storage that don't allow you to just do random overwrites. So I think that we don't really need large requests for writes. Or at least, we don't need them in the near future.

abhishek10004:
Closing this PR since we'd be going with a new approach wherein we'll only support large reads.

PR #199, opened 2026-06-29
Add support for larger reads via MaxPages config option

abhishek10004:
# Add MaxPages and MaxWrite configuration to support larger reads/writes and smaller buffers

## Description

This PR adds support for configuring `MaxPages` and `MaxWrite` in `MountConfig` (Linux only). This enables:
1. **Larger Reads**: Supporting read sizes larger than 1 MiB by increasing `MaxPages`.
2. **Larger Writes**: Supporting write sizes larger than 1 MiB by increasing `MaxWrite` (and `MaxPages` accordingly).
3. **Smaller Memory Footprint**: Reducing memory usage by configuring a smaller `MaxWrite`, which dynamically reduces the size of the pre-allocated FUSE input message buffers.

We also introduce sanitization and validation logic to ensure `MaxPages` and `MaxWrite` are compatible (i.e., `MaxPages` must be large enough to hold `MaxWrite`).

### 1. Configuration
* **`mount_config.go`**: Added `MaxPages uint16` and `MaxWrite uint32` to `MountConfig` with detailed documentation on their default behaviors and interactions.

### 2. Connection & Buffer Management
* **`connection.go`**:
  * Added `inMessageBufSize` to `Connection` to track the dynamic buffer size needed for incoming messages (`PageSize + MaxWrite`).
  * Added `sanitizeMaxPagesAndWrite()` to handle default values when one or both of `MaxPages` and `MaxWrite` are unset, and to validate that `MaxPages * PageSize >= MaxWrite`.
  * Updated `Connection.Init()` to negotiate the configured `MaxPages` and `MaxWrite` with the kernel.
* **`internal/buffer/in_message.go`**: Updated `NewInMessage()` to accept a dynamic buffer size instead of using a hardcoded constant.
* **`freelists.go`**: Updated `getInMessage()` to pass the connection's dynamic `inMessageBufSize` when allocating new message buffers.

### 3. API & Documentation Updates
* **`fuseops/ops.go`**: Updated documentation for `ReadFileOp.Dst` and `ReadFileOp.Data`:
  * Clarified that `Dst` will be `nil` when `MaxPages` is configured for reads larger than the pre-allocated buffer.
  * Clarified that the file system must use `Data` (vectored read) when `Dst` is `nil`.

### 4. Tests
* **`connection_test.go`**: Added `TestSanitizeMaxPagesAndWrite` covering various scenarios:
  * Neither option configured (defaults applied).
  * Only `MaxWrite` configured (dynamic `MaxPages` calculation).
  * Only `MaxPages` configured (dynamic `MaxWrite` calculation).
  * Both configured (valid and invalid combinations).

## Testing Done
* Added and ran `TestSanitizeMaxPagesAndWrite` in `connection_test.go`.


mount_config.go
```diff
@@ -243,6 +241,21 @@ type MountConfig struct {
...
+	// FUSE_INIT phase.
+	//
+	// If set, it must be at least the value required to support the default
+	// maximum read/write size (1 MiB), which is typically 256 pages (for 4KB
+	// page size). Configuring a value smaller than this will cause Mount to
+	// return an error.
+	//
+	// If unset, a default value based on the maximum message in/out payload
+	// sizes is used.
+	MaxPages uint16
```
vadlakondaswetha, line 265: I am assuming we will use this only for reads. can you explicitly mention that.

vadlakondaswetha (approved):

fuseops/ops.go
```diff
@@ -709,13 +709,22 @@ type ReadFileOp struct {
...
-	// The destination buffer, whose length gives the size of the read.
-	// The file system can write to this buffer for non-vectored reads.
+	// The destination buffer. If the requested read size is small enough (typically
+	// up to 1 MiB), the library provides this buffer, whose length matches the size of
+	// the read, utilizing its pre-allocated request memory. The file system can write
+	// to this buffer for non-vectored reads.
+	//
+	// If the requested read size is larger than the library's pre-allocated buffer
+	// (e.g., when MountConfig.MaxPages is configured to support larger reads), this
+	// slice will be nil.
```
geertj, line 719: Who sets this slice to nil? Does that happen here, or in GCSFuse? It might be easiest to always set this, and make it the file system's problem to interpret the GCS operation and see if there's enough space. For example, even if the requested read size is 2 MB, it's not a given that the response will be of that size.

geertj (approved):

fuseops/ops.go
abhishek10004: jacobsa/fuse sets it to nil before sending the read request to GCSFuse.

In case the read request is larger than the buffer available, jacobsa/fuse sets dst to nil. Please see GetFree function (in in_message.go) which returns nil in case we're requesting for more space than what is available. Setting it to nil means that the file system has to allocate the memory for read and return it via Data (this is documented as well).

>It might be easiest to always set this, and make it the file system's problem to interpret the GCS operation and see if there's enough space. For example, even if the requested read size is 2 MB, it's not a given that the response will be of that size.

This would not help as the file system would have to allocate a new buffer anyways if it sees that the request is larger than what dst can hold.

